PR #1, opened 2016-03-22
RT scripting improvements

rems75:
Fixed the regression, took the latest commits into account, hopefully end of this part of the project.


jkoppel:
Looks like you never addressed the security flaw. Once you've done so, we can merge this

ironfist/src/cpp/game/save.cpp
```diff
@@ -235,7 +235,7 @@ void SaveMapVariables(ironfist_map::map_t& m) {
 	GetNextMapVariable(key, mapVariableId, mapVariableValue);
 	while (mapVariableId != "noMapVariables") {
 		if (mapVariableValue == NULL) {
-			char errorMessage[50];
+			char *errorMessage = (char *)malloc(strlen(mapVariableId) + 37);
```
jkoppel: One of the big concepts in software engineering is coupling.  How much does one thing depend on another? If thing A changes, in what ways will thing B have to change?

Here, thing A are your string constants "MapVariable: " and "'could not be saved.", and thing B is the number 37. 

Imagine: You are an Ironfist player who knows some coding, and, for whatever reason, you don't like the "MapVariable cannot be saved" popup. Maybe you want to add some more information, or maybe you want to translate it? You download the codebase, search for the string "could not be saved", change it to something else.....and never notice 3 lines above there's that number 37 which needs to be changed as well. Oops.

The easy part of software engineering is getting things working. The hard part is making it hard to accidentally break things.

What's something you can do to eliminate this coupling? Think about it for a few minutes before scrolling down. I will post one possible answer in the next comment

---------------

In other news, this is a memory leak: errorMessage is allocated but never released. If you do go this route, you'll need to call free(errorMessage) when you're done using it. Actually, you should probably use the ALLOC and FREE macros rather than malloc and free, to stay consistent with the rest of the codebase. (ALLOC does some additional tracking of allocated memory to help debug memory leaks.)

jkoppel: Okay, here are a few answers:

The first: Where did that number 37 come from? Make it explicitly depend on those string constants so that it will be automatically recomputed  when they change.

char *s1 = "MapVariable '";
char *s2 = "'could not be saved.";
char *errorMessage = (char *)malloc(strlen(mapVariableId) + strlen(s1) + strlen(s2));
// rest is same

--------------

One thing that's kinda-sorta a copout is to use the global "gText" buffer instead of malloc'ing you're own. Might technically allow a security vulnerability, but it's very large, and you'd be no less insecure than most of the rest of the code, which uses gText quite often. It's also less exploitable because it's stored in a  different region of memory than a locally declared buffer, so it's harder to take over the program by overflowing it.

-----------------

The first suggestion solves the coupling problem. You may also want to use sprintf to make it easier to see what message is displayed.

sprintf(errorMessage, "%s%s%s", s1, mapVariableId, s2);

Or use snprintf to make the security more explicit:

int len = strlen(mapVariableId) + strlen(s1) + strlen(s2);
char *errorMessage = (char *)malloc(len);
snprintf(errorMessage, len, "%s%s%s", s1, mapVariableId, s2);

---------

The asprintf function would solve this problem. Unfortunately, it's not available in Visual C++. I believe it can also be done cleanly using the std:string C++ class, but I'm not very familiar with most C++ things.

PR #2, opened 2016-03-26
event loop sleep()

chrisbigart:
Trying to figure out versioning control / git(hub).
This is a small patch to heroes2_imports.inc and the addition of event_loop.cpp.  It simply hooks and then adds Sleep() to Process1WindowsMessage(), resulting in reduced CPU usage (without slowing the game down, so far as I can tell the main game loop effectively polls the windows message pump).
I don't know if any of the changes to the solution file are necessary or even useful, I'm still trying to figure out how to get everything synced up.


jkoppel:
Congratulations! You've figured out how to make a pull request.

As you can see by clicking "Files changed,"  you've included quite a few changes other than that one in this PR. You've changed and copied creatures.xml (how does that not break it?), and also modified the VS project file (which is fine, except that you included absolute paths, which could stop the game from building on someone else's machine).

If you wanted to only send a PR with the desired changes, probably the easiest way would be to create a new branch off master and then run the following two commands:

  git cherry-pick 	ee14687
  git cherry-pick  	be60d5f

(Granted, I'd probably not accept this change even if it didn't have the other stuff. Without having much stronger evidence that nothing bad would happen from adding that Sleep call, I'd be too worried about bugs to make the improved runtime worthwhile.)

chrisbigart:
Right, I didn't want to include those changes but I had to make them for things to compile on my machine.  There seems to be a couple issues building the 'master' branch unmodified relating to paths and static linking.  For a pull request, are you able to select specific files for inclusion, or is it a binary all-or-nothing thing?

I disagree with you about not adding this change until it can be verified to be bug-free.  There certainly could be bugs arising from it, but without the ability to enable it, it can't be tested and stronger evidence to support that bad things won't happen can't be produced.  If you are concerned about it breaking something, then in my opinion it needs to be added as a non-default option, and included in the official build.  That way, people (other than myself) can enable it and test it when the next release comes out.

To do that, the option has to be exposed somewhere.  What do you think is the best mechanism for this?  I could add another entry to the registry (under "SOFTWARE\\New World Computing\\Heroes of Might and Magic 2\\1.0" with the rest of the prefs) like "Sleep Event Loop" as a REG_DWORD where 0 (or -1) is a no-op, and a positive value is the number of ms to sleep.

Reducing the cpu usage isn't that big of a deal on the desktop (at least when nothing else cpu intensive is competing with it), but it can be a big deal on a mobile device where it could result in significant heat/power savings.

chrisbigart:
Okay I pared down the changed files to just the source files I changed.  I also added a couple helper functions to query this value from the registry.
By default the registry value will not exist, the function will return -1 for the value, and the call to Sleep() will not occur.  If the registry value is set, then it will call Sleep() with the reg value as an argument.

jkoppel:
Wait, how can you run Ironfist on mobile? That changes the cost/benefit analysis a bit. If CPU usage is indeed a real problem in some contexts, and this is disabled by default, then there's no reason not to include this.

Please give me details about what your build problems were; we've put effort into making sure anyone can compile it. I can say that none of the changes in the project files look like they should affect the build

A pull-request is an all or nothing thing, but only on a per-branch basis. A typical git workflow is that every new feature/bugfix/set of changes is developed in its own branch.

There are still a few problems with this PR, the main one being that the registry key is located in different locations on different computers. The Heroes 2 codebase already has some existing infrastructure for dealing with this, and Remses has just started working on a way to store Ironfist configuration settings that does not use the registry. The rest are just code quality and style problems. If you'd like, I can take over the job of making the remaining changes so that we can merge this in. (My installation of IDA is currently broken for mysterious reasons though, so I will probably need to get that fixed first.)

chrisbigart:
"Mobile device" meaning a laptop running Windows, or I suppose a tablet that runs Windows.  Or running Windows on a virtual machine from Mac/Linux, it could potentially be an issue even on a desktop depending upon the virtual machine overhead, I guess.

I will have to download and try to rebuild with the original project file because I don't remember exactly what the issues were.  I do remember that there seem to be two versions of the creatures.xml file, one will result in a parsing exception being thrown, while the other works just fine.

I'm not sure I understand what you mean about the registry key locations.  This is basically taken right from the decompilation, I don't see where the original code has this ability.

I agree that a non-registry solution to config settings would be a better choice, especially going forward.

No, I'd rather you didn't take over and make the changes.  This patch isn't so much about getting this particular feature integrated into the codebase as it is me getting adapted to the flow of git and us establishing some sort of baseline for code quality and style.  I can't read your mind on those issues, so if you could point out the problems, or better yet create a quick style guide or short-term roadmap for how you think these issues should be approached, I think that make things more efficient for future contributions.

jkoppel:
Hmmm....I run Ironfist in a Windows VM on a laptop. Is this actually a problem? *checks* 50% CPU usage of the virtual machine, dang. I guess I never notice; generally, starting my Windows VM takes up a lot of battery no matter what.

Okay, looks like I may have been mistaken about the multiple locations thing. I do recall seeing Heroes II settings in multiple different places on my machine, but I think that may have been due to differences between the original and GOG.com versions. 

I'd want to reiterate that, if I were to take this over, I would very much explain everything I'm doing. I get paid to coach people to be better software engineers, and one of the perks of working on Ironfist is getting that for free. I made the offer to do most of it for you so that you wouldn't get frustrated with getting this first change in the game, and could learn things more piece by piece rather than having to learn everything at once 

I see how you're getting this code from the decompilation. I will say that it generally takes a decent amount of work to take decompiled code and turn it into something that meets our code quality standards. I haven't figured out how much of this I could put into a syle guide, and a lot of it is fairly general code-quality stuff. I generally prefer to give this kind of feedback interactively rather than in a format like Github comments: the most straightforward approach is for me to just do a braindump of every issue I see, which can be very uncomfortable to be on the receiving end of.

So, how shall I help you?

chrisbigart:
Yeah I've noticed that there are multiple copies of HoMM2 settings in the registry, it could have to do with the GoG version but I don't think I've ever used that version and I still have multiple copies in there.  I think the issue has to do with WoW64 (Windows on Windows support for 32-bit code on 64-bit Windows), and/or some other compatibility thing Windows does.  If you look at the registry key path specified in the H2 code, and then look at the registry location where it actually exists, they differ.  On my machine the actual registry key is at `HKEY_LOCAL_MACHINE\SOFTWARE\Wow6432Node\New World Computing\Heroes of Might and Magic 2\1.0`

It doesn't matter to me, you can post it all here or we can discuss it elsewhere, whatever you think will work best.

chrisbigart:
Also I think some sort of roadmap, maybe on the wiki here, would be a good idea.
I have a number of changes in my fork, some of them are ready to go and others are still buggy or incomplete.
I have disabled the hero movement reminder, and once the preferences code is settled on, then that can be another simple option like the sleep() one.
I have CD-music (mostly) working via h2cdplay, but I need some help figuring out some details about the volume levels.  Also we can discuss how to implement reading .mp3/ogg files directly.  I have a few ideas about this.
Lastly I am working on widescreen support, and xbrz support.  I could use help on this as well.

I think collaborating on features in development would help accelerate things.

jkoppel:
Okay. I will write up my comments by EOD tomorrow.

We had a big meeting a few weeks ago where we announced our big yearly and monthly goals (also announced on the project-ironfist mailing list). With it, came a big shift in how we prioritize to be much more explicitly focused on hitting our goals (mostly about user engagement). We just did our first round of user surveys, and are only now crystallizing it into a roadmap. I'm currently talking to a mentor about what to do for tracking this kind of stuff -- volunteer projects like this can't have too much in the way of processes, so we're more limited how rigorous we can be.

I'm happy to work with you on all of those features. With the current Ironfist programmers, I do this by p by coming up with warmups to help them get better at the code, by pairing over Skype, by answering questions over chat or E-mail. I can also assist by doing some of the coding tasks and by doing some reverse-engineering work.

ironfist/src/cpp/game/event_loop.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+//adding a call to Sleep() during event processing will reduce excessive cpu usage
+//on my machine, 10ms seems to reduce cpu load to <1% while not having any effect
+//on the speed/responsiveness of the game
+
+const int MAX_SLEEP_VALUE = 1000; //sanity check
+
+void __fastcall Process1WindowsMessage_orig();
+void __fastcall Process1WindowsMessage()
+	{
```
jkoppel: This one's just a style thing:

The style I've been pushing for in the Ironfist codebase is braces on the same line as the function declaration.  I'm not familiar with any C/C++ style that calls for indenting the brace on the next line. Also, spaces over hard tabs.

I've fought with editor settings every time I upgrade Windows, so I'd be hypocritical to be too hard on this.

ironfist/src/cpp/registry_prefs.h
```diff
@@ -0,0 +1,42 @@
...
+		{
+		RegQueryValueExA(hKey, pref_key.c_str(), 0, &Type, (LPBYTE)&Data, &cbData);
+		}
+
+	return Data;
+	}
+
+template<typename T> bool write_registry_pref(const std::string& pref_key, const T& value);
+
+template<> bool write_registry_pref(const std::string& pref_key, const DWORD& value)
```
jkoppel: Why are you using std::string here? We're not using it in the rest of the codebase, and you're only using it to convert into a C string here.

ironfist/src/cpp/registry_prefs.h
```diff
@@ -0,0 +1,42 @@
...
+	{
+	DWORD cbData = 4;
+	LPCSTR SubKey = "SOFTWARE\\New World Computing\\Heroes of Might and Magic 2\\1.0";
```
jkoppel: At first I wanted to insist that you extract everything that uses this string constant from the codebase so that there's not risk of changing this but not the other locations. However, I don't think anything bad will happen if we do, so maybe not.

ironfist/src/cpp/registry_prefs.h
```diff
@@ -0,0 +1,42 @@
...
+	HKEY hKey = nullptr;
+
+	if(RegOpenKeyExA(HKEY_LOCAL_MACHINE, SubKey, 0, 0xF003Fu, &hKey) != ERROR_SUCCESS)
```
jkoppel: "0xF003Fu" is what we call a "magic number." It's not at all clear where it came from. Presumably this is a bitwise-or of several constants or a special value defined somewhere. I think it's easier to read when written as such.

jkoppel:
Okay, on second read, it's actually mostly okay, other than the formatting issues and the magic number.

ironfist/src/cpp/registry_prefs.h
chrisbigart: You're right, it's straight from the decompilation so I left it as it is.  Turns out that constant is equal to `KEY_ALL_ACCESS`. (https://msdn.microsoft.com/en-us/library/windows/desktop/ms724878(v=vs.85).aspx)

chrisbigart:
Something more like this? (I didn't fix the tabs to spaces thing)

```
#ifndef REGISTRY_PREFS_H
#define REGISTRY_PREFS_H

#include <Windows.h>
#include <string>

template<typename T> T read_registry_pref(const std::string& pref_key);

LPCSTR Heroes2RegistrySubKey = "SOFTWARE\\New World Computing\\Heroes of Might and Magic 2\\1.0";

template<> DWORD read_registry_pref(const std::string& pref_key) {
	DWORD Type;
	DWORD cbData = 4;
	DWORD Data = (DWORD)-1; 
	HKEY hKey = nullptr;
	if(RegCreateKeyA(HKEY_LOCAL_MACHINE, Heroes2RegistrySubKey, &hKey) == ERROR_SUCCESS)
		{
		RegQueryValueExA(hKey, pref_key.c_str(), 0, &Type, (LPBYTE)&Data, &cbData);
		}

	return Data;
	}

template<typename T> bool write_registry_pref(const std::string& pref_key, const T& value);

template<> bool write_registry_pref(const std::string& pref_key, const DWORD& value) {
	DWORD cbData = 4;
	HKEY hKey = nullptr;

	if(RegOpenKeyExA(HKEY_LOCAL_MACHINE, Heroes2RegistrySubKey, 0, KEY_ALL_ACCESS, &hKey) != ERROR_SUCCESS)
		return false;

	if(RegSetValueExA(hKey, pref_key.c_str(), 0, REG_DWORD, (LPBYTE)&value, cbData) != ERROR_SUCCESS)
		return false;

	return true;
	}


#endif
```

chrisbigart:
I guess it doesn't really need to be a template function, since the game seems to exclusively read REG_DWORD or REG_SZ values.

ironfist/src/cpp/registry_prefs.h
chrisbigart: C-strings are tricky, and I avoid them whenever possible.  In this case it doesn't matter, but in my opinion it would be beneficial to use them in other places in the code over C-strings, simply because there is a lot more room for error working with char*'s vs std::strings.

jkoppel:
Other than issues with braces and spacing, looks good!

.....other than that nullptr. We'd have to officially upgrade the project to C++11 to use that, and I'd have to do some research to make sure there can be no ill effects of that decision.

PR #5, opened 2016-04-03
CPU polling + Hero Movement Reminder patch

chrisbigart:
Okay, so what I did was create "read_pref" and "write_pref" in prefs.h, this is just a generic interface right now, and it does nothing but call the respective registry read/write functions.  This way, if later on we change things to use .ini or .xml files or whatever non-registry preferences approach, less code will have to be changed.

This also adds another preference, "Show Hero Movement Reminder", which works just like it does in homm3.


ironfist/src/cpp/adventure/adv.cpp
```diff
@@ -2,6 +2,57 @@
...
+				NormalDialog("One or more heroes may still move, are you sure you want to end your turn?",
+							2, -1, -1, -1, 0, -1, 0, -1, 0),
+				gpWindowManager->buttonPressedCode != 30726)
+				)
+			{
+			gpGame->NextPlayer();
+			}
+		/* //i doubt this is needed but it can be added in if it causes an issue
+		//seems to deal with updating the bottom right hero/day/kingdom view pane
+		if(if(GUIMessage_evt->yCoordOrFieldID == 4)
```
jkoppel: I'm generally against removing behavior like this until you're very sure it's okay to do so. The original Heroes II code is written in such a way that a lot of subtle bugs can occur from taking things out.

ironfist/src/cpp/adventure/adv.cpp
```diff
@@ -2,6 +2,57 @@
...
+		{
+		DWORD hero_reminder_reg_dword = read_pref<DWORD>("Show Hero Movement Reminder");
+		//default is true, but read_pref() returns -1 if the value is not set
+		bool show_hero_movement_reminder = !(hero_reminder_reg_dword == 0);
+		
+		if(!show_hero_movement_reminder || !gpCurPlayer->HasMobileHero() ||
+				(
+				NormalDialog("One or more heroes may still move, are you sure you want to end your turn?",
+							2, -1, -1, -1, 0, -1, 0, -1, 0),
+				gpWindowManager->buttonPressedCode != 30726)
```
jkoppel: The "if ((DoSomething(), condition))"  pattern is something you might see in decompiled code, but I've literally never see anyone use it in code they've written. I and probably most others don't usually don't expect effects to occur in the middle of conditionals like this, making it easier to miss things when reading this code.

jkoppel: "30726" is another example of a magic number.  You can figure out what it means by using h2guiview on whatever .bin file the NormalDialog function opens. Certainly, it's hard to know what it means just by seeing it.

We still don't have a great way of dealing with constants in the .bin GUI files, but I think we do have a few examples of doing this reasonably -- I think displaying the mage guild is one.

Yes, Tales of Enroth included a similar magic constant, and I let it into Ironfist. Shame on me.

ironfist/src/cpp/adventure/adv.cpp
jkoppel: Also, using git is supposed to remove the need for having commented-out code

jkoppel:
The new reminder is cool! In the future, the preferred workflow is to develop each feature in its own branch. In this case, you'd have the "Show Hero Movement Reminder" feature as a branch off of the branch you're developing the preferences feature in. This lets you merge them in separately.

I gave my main comments. The other thing is that there are still lots of indentation/braces problems, probably from your editor configuration.

Hope this helped!

ironfist/src/cpp/adventure/adv.cpp
chrisbigart: I'm not sure what this is really telling me:
```
//
//Detected button
//fieldID : 30726
//	field_12 : 8
//	field_14 : 2
//	flags : 11468806
//	offsetY : 106
//	width : 98
//	height : 25
//	normalImageIdx : 7
//	armedImageIdx : 8
//	field_28 : 0
//	field_2A : 1
//	icnFileID : -478270621
```

We have no way to tell what the original name was from the binary, so we just have to make up something.

```
const int BUTTON_CODE_OKAY = 30726;
```

chrisbigart:
Okay I made most of the changes you suggested. 

I'm not sure what to do about the braces/indentation; I think you should come up with a brief official style guide just for bracing/indentation, so at least I have something to look at without having to infer what style you want from existing code.  

I have difficulty reading code with the bracing/indentation style present in ironfist, but I'm sure you look at my style and think the same thing :)

chrisbigart:
Hmm actually removing the comma operator portion introduces another issue, I have to fix that first...

chrisbigart:
Okay, working as intended now.  The 
```
if ( !playerData::HasMobileHero(gpCurPlayer)
        || (NormalDialog(...), gpWindowManager->buttonPressedCode != 30726) )
```
was actually a very 'clever' or terse way of handling it, but I agree it is best to avoid it.  I had to look the `,` operator up myself because I wasn't sure what it did, like you said I don't think I've seen it in a non-example piece of code before.

jkoppel:
Okay, I've written a couple sentences on coding guidelines: http://projectironfist.pbworks.com/w/page/54023700/Getting%20Started%20with%20Coding . We're mostly following the Google C++ style code, at least for formatting: https://google.github.io/styleguide/cppguide.html#Formatting . This matches the code I've been writing personally, which is not a coincidence, as I've worked at companies with style guides influenced by Google's. 

Your style is pretty interesting, and I can definitely see how you're used to it and other things would be unreadable. I've actually never seen your style of brace indentation before. What other projects/books use it?

ironfist/src/cpp/adventure/adv.cpp
```diff
@@ -2,6 +2,62 @@
...
+#include "prefs.h"
+
+extern void __fastcall NormalDialog(char *, int, int, int, int, int, int, int, int, int);
+
+const int BUTTON_CODE_OKAY = 30725;
+const int BUTTON_CODE_CANCEL = 30726;
+
+extern long __fastcall KBTickCount();
+
+int advManager::ProcessDeSelect(struct tag_message *GUIMessage_evt, int *a3, class mapCell * *a4) {
```
jkoppel: Rename a3 and a4 to meaningful variable names. I don't have strong feelings, but we've just been using "tag_message*" or "mapCell*" instead of e.g: "class mapCell*" for parameter declarations elsewhere.

ironfist/src/cpp/adventure/adv.cpp
```diff
@@ -2,6 +2,62 @@
...
+		DWORD hero_reminder_reg_dword = read_pref<DWORD>("Show Hero Movement Reminder");
+		//default is true, but read_pref() returns -1 if the value is not set
+		bool show_hero_movement_reminder = !(hero_reminder_reg_dword == 0);
+		
+		if(gpCurPlayer->HasMobileHero()) { //if there are any heroes with movement remaining
+			if(!show_hero_movement_reminder) { //if the movement reminder is turned off, end turn
+				gpGame->NextPlayer();
+				}
+			else { //if the movement reminder is on, ask player if he/she really wants to end turn
+				NormalDialog("One or more heroes may still move, are you sure you want to end your turn?",
```
jkoppel: This should be formatted as "} else {", with the close brace at the same indentation as the matching if statement.

ironfist/src/cpp/adventure/adv.cpp
```diff
@@ -2,6 +2,62 @@
...
+								2, -1, -1, -1, 0, -1, 0, -1, 0);
+
+				if(gpWindowManager->buttonPressedCode != BUTTON_CODE_CANCEL)
+					gpGame->NextPlayer();
+				}
+			}
+		else { //there are no heroes with movement points left, end turn
+			gpGame->NextPlayer();
+			}
```
jkoppel: Similarly, this close brace goes at the same indentation as the matching else statement.

ironfist/src/cpp/adventure/adv.cpp
```diff
@@ -2,6 +2,62 @@
...
+				}
+			else {
+				giBottomViewOverride = 2;
+				}
+			giBottomViewOverrideEndTime = KBTickCount() + 3000;
+			UpdBottomView(1, 1, 1);
+		}
+		return 1;
+	}
+	return ProcessDeSelect_orig(GUIMessage_evt, a3, a4);
```
jkoppel: Maybe put this in an else block so it's more clear this is never hit if a different button is hit?

ironfist/src/cpp/adventure/adv.cpp
jkoppel: Also, lowercase names for variables.

ironfist/src/cpp/adventure/adv.cpp
```diff
@@ -2,6 +2,62 @@
...
+const int BUTTON_CODE_CANCEL = 30726;
+
+extern long __fastcall KBTickCount();
+
+int advManager::ProcessDeSelect(struct tag_message *GUIMessage_evt, int *a3, class mapCell * *a4) {
+	extern int giBottomViewOverride;
+	extern int iCurBottomView;
+	extern int giBottomViewOverrideEndTime;
+	
+	if(GUIMessage_evt->yCoordOrFieldID == 4) {
```
jkoppel: This is one magic number to definitely remove -- presumably, 4 is the "end-turn" button. You can use h2guiview to find out for sure.

ironfist/src/cpp/adventure/adv.cpp
```diff
@@ -2,6 +2,62 @@
...
+			else if(giBottomViewOverride) {
+				giBottomViewOverride = 0;
+				}
+			else if(iCurBottomView == 2) {
+				giBottomViewOverride = 1;
+				}
+			else {
+				giBottomViewOverride = 2;
+				}
+			giBottomViewOverrideEndTime = KBTickCount() + 3000;
```
jkoppel: This block here is still quasi decompiled code. Probably at least put a comment in that its importance is unknown.  Ideally, we'd figure out exactly what it's for and replace all constants with something meaningful. I think I'd be okay with letting this in as is + a comment though because it's still fairly readable.

ironfist/src/cpp/adventure/adv.cpp
jkoppel: Yeah, h2guiview is not the best.

You should see another window where it's displaying the actual GUI. Seeing whatever is drawn at that y coordinate, width, and height tells you what it is.

There's a hash function in resources.cpp that converts filenames into an ID. It's saying that, whatever ICN file has that ID, it's the 7th and 8th images in that ICN.

Or, you can just

Also, if you dump the .rsrc section of HEROES2W.exe using PE Explorer, it will tell you all the GUID codes for things like e.g.: the File menu. I believe they like to share codes between that and the actual game GUI.

chrisbigart:
If you want to make these changes and push it through, that's fine with me.

I have the CD music code working, but it needs some cleanup and beautification.  I don't know when I'm going to get around to that, and I could use some additional eyes / contributions from others on the code anyway.  What if you were to create a branch for the CD music patches, then I could create a pull request to that branch with the current progress, that way the unfinished code isn't in the master branch yet?  Or would there be a better way to do things?

chrisbigart:
Also I have some issues with the xbrz/widescreen work that I want to have a discussion about.  What's the best way to do that?  Open up an issue, or create a page on the wiki?  Or the mailing list?

jkoppel:
Okay. I'll go make the final changes to get this merged in.

For the CD music code, go ahead and create a pull request like that just for review purposes. That's actually a pretty common practice.

For the xbrz questions, probably start a thread on the mailing list.

jkoppel:
Tested and merged in! Congratulations, you're an Ironfist contributor. :)

I actually had some trouble during testing because I had Heroes 2 data in two different paths in my registry, and I modified the wrong one at first.

PR #6, opened 2016-04-14
CD Music changes

chrisbigart:
This is a mess, but it does work.  This PR is just for everyone else to take a look at it, I guess, and create/propose some changes to clean things up.

It loads .mp3 files of the tracks from the (default) "tracks" directory, labeled as "track00.mp3" through "track43.mp3".


jkoppel:
I want a cleaned up version of this in the next Ironfist release.

I'm just going to take notes here on what's left to be done.

1) Work should be done in new branch, with only relevant changes included. Changes to sound.cpp, sound.h, and inclusion of BASS
2) BASS needs to be added to installer
3) soundManager class definition needs to be minimized. Should extend baseManager; only fields with known names should be labeled
4) Fields of soundManager that are used should be reverse-engineered and named

jkoppel:
5) Code should be well factored and converted to follow the Ironfist style guidelines
6) State should be properly encapsulated, probably by creating a BassState structure and placing it at the end of the soundManager object
7) Must be able to play music however it's shipped with the GOG.com version
8) Parameter names and purposes of soundManager methods must be reverse-engineered
9) Additional soundManager reverse engineering needs to be done to ensure we have our corners covered

jkoppel:
Looking this over, this is a fair bit of work, and I don't think we have someone ready to take it on right now (at least not if we want to cover #9). I think we'll have to cut this from the 1.0 release.

jkoppel:
Moving these notes to Trello

PR #9, opened 2016-04-30
Lots of scripting additions

jkoppel:
@zhxsxuan @rems75 @chrisbigart I'd like someone to have a look at this before merging in.


ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -28,374 +31,773 @@ char *script_contents = NULL;
...
+  char* qst = (char*)luaL_checkstring(L, 1);
+  lua_pushboolean(L, H2QuestionBox(qst));
+  return 1;
+}
+
+int l_inputBox(lua_State *L) {
+  char* qst = (char*)luaL_checkstring(L, 1);
+  int len = (int)luaL_checknumber(L, 2);
+  char* input = H2InputBox(qst, len);
+  lua_pushstring(L, input);
```
chrisbigart:
```
void lua_pushstring (lua_State *L, const char *s);
Pushes the zero-terminated string pointed to by s onto the stack. Lua makes (or reuses) an internal copy of the given string, so the memory at s can be freed or reused immediately after the function returns. The string cannot contain embedded zeros; it is assumed to end at the first zero.
```
would this mean 'input' is going to get leaked?

jkoppel: Thanks. Fixed.

How much of this have you looked over?

chrisbigart:
I don't know enough about lua and the lua-c interface to really comment on most of it.  My only concern would be code like: 
```
+  hero *hro = (hero*)lua_touserdata(L, 1);
 +  int x = (int)luaL_checknumber(L, 2);
 +  int y = (int)luaL_checknumber(L, 3);
 +  gpAdvManager->TeleportTo(hro, x, y, 0, 0);
```
None of these values seem to be checked, is there some sort of sanitization occurring higher up?  If not, typos in scripts will cause the game to (best case) crash, or silently corrupt the game state.

jkoppel:
I don't have a good story currently for how to properly handle out-of-range values.

My current policy is to punt on this. I've created a Trello card to figure this out.

jkoppel:
@rems75 @zhxsxuan Review really appreciated

PR #11, opened 2016-05-05
XP compatibility

Kert:
It didn't make a lot of sense that such an old game required a newer OS like win7 to run ironfist mod.

### DIFF
--- a/ironfist/src/cpp/game/save.cpp
+++ b/ironfist/src/cpp/game/save.cpp
@@ -237,9 +237,9 @@ void SaveMapVariables(ironfist_map::map_t& m) {
 		if (mapVariableValue == NULL) {
 			char *s1 = "MapVariable '";
 			char *s2 = "' could not be saved.";
-			int len = strlen(mapVariableId) + strlen(s1) + strlen(s2);
+			int len = strlen(mapVariableId) + strlen(s1) + strlen(s2) + 1;
 			char *errorMessage = (char *)ALLOC(len);
-			_snprintf_s(errorMessage, sizeof(errorMessage), len, "%s%s%s", s1, mapVariableId, s2);
+			_snprintf_s(errorMessage, len * sizeof(errorMessage), len, "%s%s%s", s1, mapVariableId, s2);
 			DisplayError((const char*) errorMessage, "mapVariable Error");
 			FREE(errorMessage);
 		}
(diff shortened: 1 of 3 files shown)


ironfist/src/cpp/game/save.cpp
```diff
@@ -239,7 +239,7 @@ void SaveMapVariables(ironfist_map::map_t& m) {
 			char *s2 = "' could not be saved.";
 			int len = strlen(mapVariableId) + strlen(s1) + strlen(s2);
 			char *errorMessage = (char *)ALLOC(len);
-			snprintf(errorMessage, len, "%s%s%s", s1, mapVariableId, s2);
```
jkoppel: So, snprintf wasn't added till later?

Nonetheless, this is a buffer overflow security vulnerability which would allow a malicious mapmaker to take over user's computers. Not something we want. You can use sprintf_s

ironfist/src/cpp/game/save.cpp
```diff
@@ -239,7 +239,7 @@ void SaveMapVariables(ironfist_map::map_t& m) {
 			char *s2 = "' could not be saved.";
 			int len = strlen(mapVariableId) + strlen(s1) + strlen(s2);
 			char *errorMessage = (char *)ALLOC(len);
-			snprintf(errorMessage, len, "%s%s%s", s1, mapVariableId, s2);
+			_snprintf_s(errorMessage, sizeof(errorMessage), len, "%s%s%s", s1, mapVariableId, s2);
```
jkoppel: @Kert 

First, Github doesn't automatically notify people when you add commits to a pull request. In the future, please mention this in the Skype channel, or add a comment and tag me.

This is almost ready, but there's still a bug. sizeof(errorMessage) will actually be equal to 4, not len. "sizeof" is always a compile-time constant.

While we're at it, the calculation of "len" on line 240 is wrong -- there should be an extra +1 for the null terminator.

Kert:
@jkoppel Alright.
Why I forgot to multiply by length - I have no idea. I know about the sizeof~
It's fixed now I think.

jkoppel:
I thought a bit about what's stopping us from using STL strings everywhere, and it's less than I thought. You should have just gotten my E-mail with a proposal for how we'll manage the switch.

In the meantime.....I think you just want len, as opposed to len*4 which is what you have now. Or did you mean sizeof(*errorMessage), not sizeof(errorMessage)?

Kert:
@jkoppel Yes, it's a typo. I'm leaving this C-styled for the time being because refactoring unrelated code will make little sense in this branch.

jkoppel:
Good stuff.

PR #13, opened 2016-05-14
Rt activity monitoring

rems75:


ironfist/src/cpp/analytics.cpp
```diff
@@ -3,52 +3,44 @@
...
 	if (!getUuid) {
-		UUID uuid;
-		UuidCreate(&uuid);
-		unsigned char * str;
-		UuidToStringA(&uuid, &str);
-		std::string s((char*)str);
-		RpcStringFreeA(&str);
-		sUuid = s;
-		bool success = write_pref("Uuid", sUuid);
+		Poco::UUID pocoUuid = Poco::UUIDGenerator().create();
```
jkoppel: Nice!

ironfist/src/cpp/analytics.cpp
```diff
@@ -11,6 +11,10 @@
 // For UUID handling
 #pragma comment(lib, "rpcrt4.lib") 
 
+extern std::string gameAction = "gameAction";
```
jkoppel: Can you make these const?

ironfist/src/cpp/analytics.h
```diff
@@ -1,3 +1,7 @@
 #include <string>
 
-void send_event(std::string category, std::string action);
+extern std::string gameAction;
+extern std::string mapAction;
+extern std::string open;
+
+void send_event(std::string category, std::string action);
```
jkoppel: Make these references so that you don't need to copy the parameter to pass in it. Make these const to enforce that this method will not change the values.

PR #14, opened 2016-05-18
Rt array map variables

rems75:


ironfist/src/xsd/map_xml.xsd
```diff
@@ -83,6 +83,8 @@
   <xs:complexType name="mapVariable_t">
     <xs:attribute name="id" type="xs:string"/>
     <xs:attribute name="value" type="xs:string"/>
+    <xs:attribute name="isTable" type="xs:boolean"/>
```
jkoppel: Will this break old saves?

ironfist/src/cpp/gui/dialog.cpp
```diff
@@ -22,4 +23,14 @@ void DisplayError(const char* msg, const char* title) {
 	MessageBox(NULL, (LPCWSTR)wideCharMsg, (LPCWSTR)wideCharTitle, MB_OK);
 	FREE(wideCharMsg);
 	FREE(wideCharTitle);
-}
\ No newline at end of file
+}
+
+void ErrorSavingMapVariable(const char* mapVariableId) {
```
jkoppel: We're adopting the std:string changes! Applies to all future code being merged in, this included :)

ironfist/src/cpp/gui/dialog.cpp
```diff
@@ -26,11 +27,10 @@ void DisplayError(const char* msg, const char* title) {
...
-	char *errorMessage = (char *)ALLOC(len);
-	_snprintf_s(errorMessage, len * sizeof(*errorMessage), len, "%s%s%s", s1, mapVariableId, s2);
-	DisplayError((const char*)errorMessage, "mapVariable Error");
-	FREE(errorMessage);
+	const std::string s1("MapVariable '");
+	const std::string s2("' could not be saved properly.");
+	const std::string mapVariableIdString(mapVariableId);
+	const std::string errorMessage = s1 + mapVariableIdString + s2;
+	const std::string errorLabel("mapVariable Error");
+	DisplayError(errorMessage.c_str(), errorLabel.c_str());
```
jkoppel: Please create an overloaded version of DisplayError which accepts std:string& so you don't have to do this.

rems75:
Upgraded version, taking into account lua types.

ironfist/installer/map_xml.xsd
```diff
@@ -81,14 +81,24 @@
   </xs:complexType>
 
   <xs:complexType name="mapVariable_t">
+    <xs:sequence>
+      <xs:element name="array" minOccurs="0" maxOccurs="unbounded">
```
jkoppel: So, "arrays" and "maps" are actually just "tables" in Lua. Should name accordingly; map variables are fundamentally Lua values.

ironfist/installer/map_xml.xsd
```diff
@@ -81,14 +81,24 @@
   </xs:complexType>
 
   <xs:complexType name="mapVariable_t">
+    <xs:sequence>
+      <xs:element name="array" minOccurs="0" maxOccurs="unbounded">
+        <xs:complexType>
+          <xs:attribute name="key" type="xs:string"/>
+          <xs:attribute name="value" type="xs:string"/>
```
jkoppel: This needs to be recursive. You can have tables be values in tables. Important for having e.g.: 2D arrays.

ironfist/src/cpp/game/save.cpp
```diff
@@ -228,28 +228,38 @@ ironfist_map::hero_t WriteHeroXML(hero* hro) {
 	return hx;
 }
 
+void SetMapVariableValue(ironfist_map::mapVariable_t &mapVar, luaTable lt) {
+	for (luaTable::const_iterator it = lt.begin(); it != lt.end(); ++it) {
```
jkoppel: I could tell this code was better than previous in an instant.

Notice the simple control flow, and how the control flow mirrors the structure of the data. When you update the map_xml so that tables can be recursive, this will correspondingly become a Lua function.

ironfist/src/cpp/gui/dialog.cpp
```diff
@@ -24,17 +25,27 @@ char* H2InputBox(char* msg, int len) {
...
-	wchar_t* wideCharMsg = ConvertNarrowToWide(msg);
-	wchar_t* wideCharTitle = ConvertNarrowToWide(title);
-	MessageBox(NULL, (LPCWSTR)wideCharMsg, (LPCWSTR)wideCharTitle, MB_OK);
-	FREE(wideCharMsg);
-	FREE(wideCharTitle);
+void DisplayError(std::string msg, std::string title) {
+	MessageBoxA(NULL, msg.c_str(), title.c_str(), MB_OK);
+}
+
+void ErrorMapVariable(std::string mapVariableId, const std::string s2, std::string addErrorMessage) {
```
jkoppel: dialog.cpp contains general utilities for displaying messages. There should be no details  These should all be put in save.cpp.

I think I'd like to spend our next session exploring the principle behind this.

ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -772,38 +772,116 @@ char *GetScriptContents() {
...
-  }
-  else {
-    mapVariableId = "noMapVariables";
-  }
-}
-
-void SetMapVariables(const char *id, const char *quantity) {
-  lua_pushstring(map_lua, quantity);
-  lua_setglobal(map_lua, id);
+bool isTable(std::string valueType) {
```
jkoppel: It makes perfect sense locally to ask if a string is a table, but not generally. Make this function "static" to emphasize this.

Most of the codebase doesn't do this, but I'd like to adopt this convention more going forward. I'm putting a not in the style guidelines.

Also, please have it accept a string& instead of a raw string. The way you have it now, the value will be copied every time you invoke this function.

ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -772,38 +772,116 @@ char *GetScriptContents() {
...
+	return (valueType == "table");
+}
+
+bool isStringNumBool(std::string valueType) {
```
jkoppel: Same as above.

ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -772,38 +772,116 @@ char *GetScriptContents() {
...
+	return (valueType == "string" || valueType == "number" || valueType == "boolean");
+}
+
+std::string GetValue(std::string mapVariableType) {
+	if (mapVariableType == "string") {
+		std::string stringValue(lua_tostring(map_lua, -1));
+		return stringValue;
+	}
+	else if (mapVariableType == "number") {
```
jkoppel: Format this as "} else if (" ...

ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -772,38 +772,116 @@ char *GetScriptContents() {
...
-  lua_setglobal(map_lua, id);
+bool isTable(std::string valueType) {
+	return (valueType == "table");
+}
+
+bool isStringNumBool(std::string valueType) {
+	return (valueType == "string" || valueType == "number" || valueType == "boolean");
+}
+
+std::string GetValue(std::string mapVariableType) {
```
jkoppel: Get value of what? Probably should have a name like "MapVariableGetValue" or "GetMapVariableValue".

ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -772,38 +772,116 @@ char *GetScriptContents() {
...
+	if (mapVariableType == "string") {
```
jkoppel: I expect Lua declares these strings as constants somewhere.

ironfist/src/cpp/scripting/scripting.h
```diff
@@ -1,6 +1,16 @@
 #ifndef SCRIPTING_H
 #define SCRIPTING_H
 
+#include <map>
+
+typedef std::map<std::string, std::pair<std::string, std::string>> luaTable;
+
+struct mapVariable {
+	std::string luaType;
+	std::string singleValue;
```
jkoppel: The way you have this now, every mapVariable has space for both a singleValue and a tableValue.

More fundamentally, this type definition is telling you a mapVariable has both a singleValue and a tableValue, when it only has one.

Also, the strings for the Lua type don't really tell you what should be in that field; you have to keep doing string comparisons.

Here's how I'd design this data structure:

enum MapVarType { MAPVAR_TYPE_STRING, MAPVAR_TYPE_INT, MAPVAR_TYPE_TABLE, ... };

struct mapVariable {
  MapVarType type;
  union {
    long long intValue; // I think ints in Lua are 64-bit; you can check;
    double floatValue;
    luaTable tableValue;
   ...
  } value;
}

ironfist/xml/map_xml.xsd
```diff
@@ -80,17 +79,44 @@
...
+    <xs:attribute name="key" type="xs:string"/>
+    <xs:attribute name="value" type="xs:string"/>
+    <xs:attribute name="type" type="xs:string" default="string"/>
+  </xs:complexType>
+
+  <xs:complexType name="table_t">
+    <xs:sequence>
+      <xs:element name="tableId" type="xs:ID"/>
+      <xs:element name="tableElement" type="tableElement_t" minOccurs="0" maxOccurs="unbounded"/>
+      <xs:element name="parentTable" type="xs:IDREF" minOccurs="0" maxOccurs="1"/>
```
jkoppel: This seems unnecessary

ironfist/src/cpp/game/save.cpp
```diff
@@ -228,28 +228,63 @@ ironfist_map::hero_t WriteHeroXML(hero* hro) {
 	return hx;
 }
 
+std::string MapVarTypeToString(MapVarType type) {
+	if (type == MAPVAR_TYPE_STRING) {
+		return "string";
+	} else if (type == MAPVAR_TYPE_NUMBER) {
+		return "number";
```
jkoppel: Is there no good way to distinguish between floats and ints?

ironfist/xml/map_xml.xsd
```diff
@@ -80,17 +79,44 @@
...
+    </xs:sequence>
+  </xs:complexType>
+
+  <xs:complexType name="mapVariable_t">
+    <xs:sequence>
+      <xs:element name="table" type="table_t" minOccurs="0" maxOccurs="unbounded"/>
```
jkoppel: The way it's stored should mirror the actual meaning unless there's a pressing reason not to. A mapVar is defined as being either a table or a primitive type, so I'd expect to see a use of xs:choice in here.

ironfist/src/cpp/game/save.cpp
rems75: There is no integer type in Lua. I guess we could compare a number to its floor, but what would be the necessity behind it?

ironfist/xml/map_xml.xsd
rems75: The issue here is not to break old saves. The attributes value and id have to stay the same. 
As far as recursiveness is concerned, the way I designed it a table has elements and subtables (defined by the IDs of those subtables). I guess it is not the purest of recursions, I'll try to change that.

ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -772,41 +772,163 @@ char *GetScriptContents() {
...
+		std::string mapVariableId(lua_tostring(map_lua, -1));
+		lua_getglobal(map_lua, mapVariableId.c_str());
+		MapVarType mapVariableType = StringToMapVarType(lua_typename(map_lua, lua_type(map_lua, -1)));
+		mapVariable mapVar;
+		mapVar.type = mapVariableType;
+		if (isTable(mapVariableType)) {
+			luaTable lt;
+			GetMVTablesFromLUA(lt, mapVariableId);
+			mapVar.tableValue = lt;
+		}
```
jkoppel: Brace problem

ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -772,41 +772,163 @@ char *GetScriptContents() {
...
+	const std::string errorLabel("mapVariable Error");
+	DisplayError(errorMessage, errorLabel);
+}
+
+void ErrorSavingMapVariable(std::string& mapVariableId, const std::string& addErrorMessage) {
+	const std::string s2("' could not be saved properly.");
+	ErrorMapVariable(mapVariableId, s2, addErrorMessage);
+}
+
+void ErrorLoadingMapVariable(std::string& mapVariableId, const std::string& addErrorMessage) {
```
jkoppel: Brace problem

rems75: What is the problem here?

jkoppel: Hmm. My suspicion is that my screen width was such that the brace got displayed on the next line, and I didn't realize it was actually on the same line. This is fine.

ironfist/src/cpp/game/save.cpp
```diff
@@ -228,28 +247,53 @@ ironfist_map::hero_t WriteHeroXML(hero* hro) {
...
+
+	for (luaSubTables::const_iterator it = lte_lst.second.begin(); it != lte_lst.second.end(); ++it) {
+		xsdTable.subTable().push_back(WriteMapVariableTableXML(*it, lt));
+	}
+	return xsdTable;
+}
+
+void WriteMapVariablesXML(ironfist_map::map_t& m) {
+
+	std::map<std::string, mapVariable> mapVariables = LoadMapVariablesFromLUA();
```
rems75: The mapvariables are build inside that function, and contain pointers to strings and luatables. I think the data gets destructed when that function exits.

ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -772,41 +772,156 @@ char *GetScriptContents() {
...
+			ErrorSavingMapVariable(mapVariableId, " Wrong type in the table.");
+		}
+	}
+	lt[mapVariableId] = std::pair<luaTableElements, luaSubTables>(lte, lst);
+	lua_pop(map_lua, 1);
+}
+
+std::map<std::string, mapVariable> LoadMapVariablesFromLUA() {
+
+	std::map<std::string, mapVariable> mapVariables;
```
jkoppel: Note that this is a local variable. That means that, when you call some function, it allocates some space for this variable, which will be removed when the function returns.

This is fine because you are returning a map, not a pointer. That means that this map will actually get copied when it's returned. (A good compiler will eliminate the copy.)

ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -772,41 +772,156 @@ char *GetScriptContents() {
...
+	lua_getglobal(map_lua, "mapVariables");
+
+	if (lua_isnil(map_lua, -1)) return mapVariables;
+
+	lua_pushnil(map_lua);
+	while (lua_next(map_lua, -2) != 0) {
+		std::string mapVariableId(lua_tostring(map_lua, -1));
+		lua_getglobal(map_lua, mapVariableId.c_str());
+		MapVarType mapVariableType = StringToMapVarType(lua_typename(map_lua, lua_type(map_lua, -1)));
+		mapVariable mapVar;
```
jkoppel: Note that this is a local variable within the while loop. That means that each version of mapVar for each iteration of the loop will be the same. If you evaluated &mapVar somewhere, that value would be the same for each iteration of the loop.

All this is fine because the mapVariables map stores a mapVariable, not a mapVariable* . That means that, when you write into mapVariables on line 881, it will actually be copied into the map.

ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -772,41 +772,156 @@ char *GetScriptContents() {
...
+		mapVar.type = mapVariableType;
+		if (isTable(mapVariableType)) {
+			luaTable lt;
```
jkoppel: This is a local variable within the while loop. That means it's going to use the same memory over and over for this variable. Each mapVar will have the same tableValue, and, after the function returns, it will point to garbage.

ironfist/src/cpp/scripting/scripting.cpp
```diff
@@ -772,41 +772,156 @@ char *GetScriptContents() {
...
+			GetMVTablesFromLUA(lt, mapVariableId);
+			mapVar.tableValue = &lt;
+		} else if (isStringNumBool(mapVariableType)) {
+			mapVar.singleValue = &GetMVValueFromLUA(mapVariableType);
```
jkoppel: This is also a problem. GetMVValueFromLua returns a string which is stored locally by the function, but it's the same value for each iteration of the loop; same problem as with lt above.

ironfist/src/cpp/game/save.cpp
```diff
@@ -456,21 +456,22 @@ void game::LoadGame(char* filnam, int newGame, int a3) {
 			std::map<std::string, mapVariable> mapVariables;
 			for (ironfist_map::map_t::mapVariable_const_iterator it = mp->mapVariable().begin();
 				it != mp->mapVariable().end(); it++) {
-				mapVariable mapVar;
+				mapVariable *mapVar = new mapVariable;
```
rems75: Do we need to add a destructor somewhere for that object? Or does the map destructor for mapVariables (3 lines above) do that for us?

ironfist/src/cpp/game/save.cpp
```diff
@@ -406,18 +447,34 @@ void game::LoadGame(char* filnam, int newGame, int a3) {
...
+					std::string *sV = new std::string;
+					*sV = it->value().get();
+					mapVar->singleValue = sV;
+				} else {
+					ErrorLoadingMapVariable(mapVariableId, " A map variable can only be a table, number, string or boolean.");
+				}
+				mapVariables[mapVariableId] = *mapVar;
 			}
+			WriteMapVariablesToLUA(mapVariables);
+
```
rems75: This is where loading crashes. 
Just before reaching this line, the program calls the map variable destructor for all the map variables in the list mapVariables.
But then, right at the next line, it calls the destructor again on a destroyed object, which raises the exception
     "Exception thrown: read access violation.
      _Parent_proxy was 0xDDDDDDDD."
I found a possibly relevant piece of information about that problem: the variable "was already deleted before but was not set to NULL afterwards".

jkoppel: To clarify: Is it still crashing after these changes?

Because you are allocating the variables with new instead of on the stack, the destructor should not be called until they are explicitly deleted.

ironfist/src/cpp/game/save.cpp
jkoppel: I have researched this issue.

If we were using C++11, we could use std::uniq_ptr, which would let us solve this problem.

After we get Sergei's build issues resolved, I'd like to take a look at officially adopting C++11 for the project (well, at least part of it: VS2015 only partially supports C++11). But.....that hasn't happened yet.

It seems you'll need to explicitly loop through the variables and delete them. And do similarly for the destructor  in scripting.h .

ironfist/src/cpp/scripting/scripting.h
```diff
@@ -4,16 +4,32 @@
...
+  MAPVAR_TYPE_STRING,
+  MAPVAR_TYPE_NUMBER,
+  MAPVAR_TYPE_TABLE,
+  MAPVAR_TYPE_BOOLEAN,
+  MAPVAR_TYPE_ERROR
+};
 
 typedef std::map<std::string, std::pair<MapVarType, std::string>> luaTableElements;
 typedef std::list<std::string> luaSubTables;
 typedef std::map<std::string, std::pair<luaTableElements, luaSubTables>> luaTable;
```
jkoppel: @rems75 

Remind me why this current representation of tables. I don't find this intuitive.

ironfist/src/cpp/game/save.cpp
rems75: Yes, this is where it crashes. It seems like the mapVariable destructor is implicitely called on line 475 when exiting the try.

jkoppel: Put a breakpoint in the destructor.

It makes sense that mapVariables itself is destructed when exiting the try. I am not expecting it to  delete its members. And I am definitely not expecting them to be deleted earlier.

ironfist/src/cpp/scripting/scripting.h
```diff
@@ -6,19 +6,18 @@
...
+	MAPVAR_TYPE_ERROR
 };
 
-typedef std::map<std::string, std::pair<MapVarType, std::string>> luaTableElements;
-typedef std::list<std::string> luaSubTables;
-typedef std::map<std::string, std::pair<luaTableElements, luaSubTables>> luaTable;
+struct mapVariable;
 
-struct mapVariable {
+typedef std::map<std::string, mapVariable> luaTable;
```
rems75: So simple

jkoppel:
Beautiful! This is how code should look.

ironfist/src/cpp/game/save.cpp
```diff
@@ -144,23 +144,27 @@ void ReadHeroXML(ironfist_map::hero_t& hx, hero* hro) {
...
+		mapVariable *mapVar = new mapVariable;
+		mapVar->type = StringToMapVarType(tableE->type());
 
-	for (ironfist_map::table_t::subTable_const_iterator subTable = table.subTable().begin();
-		subTable != table.subTable().end(); subTable++) {
-		lst.push_back(subTable->tableId().get());
-		ReadMapVarXML(lt, *subTable);
+		std::string *sV = new std::string(tableE->value().get());
+		mapVar->singleValue = sV;
+		(*lt)[tableE->key().get()] = *mapVar;
```
jkoppel: This looks like a memory leak. You're allocating new memory for mapVar, copying it into the table, and....the pointer stays around.

jkoppel:
Having trouble commenting on the diff, so commenting here:

* Why is DisplayError exported?

From the coding guidelines: "Functions which are not intended for use outside the file where they are defined should be marked static." (Yes, this requirement is somewhat new.)

jkoppel:
Line 877 of scripting.cpp has a misformatted brace

jkoppel:
After this, we should finally be done :)

rems75:
Actually DisplayError is used in both save.cpp and scripting.cpp (mostly to do checks about data structure).
I wouldn’t mind you checking them, as it’s the kind of thing I’m undoubtedly lacking at.

Le 28 juil. 2016 à 23:03, James Koppel <notifications@github.com<mailto:notifications@github.com>> a écrit :


Having trouble commenting on the diff, so commenting here:

  *   Why is DisplayError exported?

From the coding guidelines: "Functions which are not intended for use outside the file where they are defined should be marked static." (Yes, this requirement is somewhat new.)

ironfist/src/cpp/game/save.cpp
rems75: Makes kinda sense. But that mapVar becomes part of a luatable, which belongs to a "bigger" mapVariable, which belongs to a list of map variables.
From what I understand, said list is deleted when exiting try (line 476). I assume all elements of the list get destructed sequentially, which includes the map variable containing the luatable that contains the problematic mapVar.
How deep is the destruction performed by delete luaTable?
More generally, what tools can I use to check if there is a memory leak or not (I added a breakpoint in the mapVariable destructor, it is called a bunch of times, but it's tough to be completely sure)?

jkoppel:
So, first, DisplayError in scripting.cpp is confusingly named, because it implies that it's a general capability, but it's scripting-specific. It also can easily be confused with the general DisplayError function defined in dialog.h . Name it something like ScriptError() instead?

It also looks like it's only used within scripting.cpp, so it can be removed from scripting.h, and declared static.

So, here's what remains:

* The above naming issue
* Marking functions static
* Any remaining testing
* Merging into master. This will be nontrivial because of the line ending changes. I just made a commit that did most of this for you.

PR #16, opened 2016-05-23
Hz creature secondary cost

zhxsxuan:
@jkoppel Ready for review.


ironfist/src/cpp/combat/creatures.cpp
```diff
@@ -50,6 +52,17 @@ attributeNameTableEntry creatureAttributeNameTable[] = {
 	{"undead", UNDEAD}
 };
 
+struct secondaryCostIdxTableEntry { char * resource; int idx; };
+
+secondaryCostIdxTableEntry secondaryCostIdxTable[] = {
+	{"wood", 0},
```
jkoppel: These integer constants are "magic numbers."

In the words of my TA, "The number 1 is not mercury. The number  is the number 1."

I'd like you to cultivate a habit of always looking for a place where these constants are already defined. In this case, the file HEROES2W.h taken from the reverse-engineering work defines an enum called RESOURCES with constants like RESOURCE_WOOD, etc. If this isn't already somewhere in the Ironfist project, you can add it, and then refer to those constants.

I think there should also be a table somewhere in the executable giving the names of resources by ID. You should reuse that table. Like, instead of having an entry be {"wood", 0}, have the entry be {gResourceNames[RESOURCE_WOOD], RESOURCE_WOOD}.

ironfist/src/cpp/combat/creatures.cpp
```diff
@@ -137,6 +150,31 @@ void LoadCreatures() {
...
+				}
+				bool secondary_cost = false;
+				for (creature_t::secondary_cost_iterator i = c.secondary_cost().begin();
+				     i != c.secondary_cost().end();
+					 ++i) {
+					    for (int k = 0; k < ELEMENTS_IN(secondaryCostIdxTable); k++) {
+						   if (strcmp(secondaryCostIdxTable[k].resource, i->resource().c_str())
+							   == 0) {
+							   if (!secondary_cost) {
+								   gMonSecondaryCost[id][k][0] = 1;
```
jkoppel: gMonSecondary cost should probably just have dimension [MAX_CREATURES][NUM_SECONDARY_RESOURCES]. That 0/1 field is redundant (I'll explain below how to eliminate the only place it's used).

And yeah, having some constant somewhere like "NUM_SECONDARY_RESOURCES" is much preferred to having the magic number "6" floating around.

ironfist/src/cpp/combat/creatures.cpp
```diff
@@ -137,6 +150,31 @@ void LoadCreatures() {
...
+								   gMonSecondaryCost[id][k][1] = i->cost();
+								   secondary_cost = true;
+							   }
+							   else {
+								   EarlyShutdown("Startup Error", 
+									   "Error loading creatures.xml.\nToo many secondary resource costs per creature!");
```
jkoppel: Oh cool; EarlyShutdown is definitely  a better choice than DisplayError. Thanks!

It would be nice to also give the name of the creature. We've adopted the new std::string requirements, so this would mean using std::string for the string manipulation, and overloading EarlyShutdown to also accept an std::string .

ironfist/src/cpp/combat/creatures.cpp
```diff
@@ -167,4 +205,51 @@ void UnloadCreatures() {
...
+		{
+			costs[i] = gMonSecondaryCost[mon][i][1];
+			customed_sc = true;
+		}
+		else
+			costs[i] = 0;
+	}
+	costs[6] = gMonsterDatabase[mon].cost;
+
+	if (!customed_sc) {
```
jkoppel: Just put this information into creatures.xml , and throw away all this logic.

ironfist/src/cpp/combat/creatures.cpp
```diff
@@ -167,4 +205,51 @@ void UnloadCreatures() {
...
+	{
+		if (gMonSecondaryCost[mon][i][0])
+		{
+			costs[i] = gMonSecondaryCost[mon][i][1];
+			customed_sc = true;
+		}
+		else
+			costs[i] = 0;
+	}
+	costs[6] = gMonsterDatabase[mon].cost;
```
jkoppel: Magic number.

ironfist/src/cpp/combat/creatures.cpp
```diff
@@ -167,4 +205,51 @@ void UnloadCreatures() {
...
 	}
+}
+
+void __fastcall GetMonsterCost(int mon, int *const costs) {
+	
+	int i;
+	bool customed_sc = false;
+	for (i = 0; i < 6; ++i)
+	{
+		if (gMonSecondaryCost[mon][i][0])
```
jkoppel: Once you get rid of customed_sc, this entire if-else will be equivalent to "costs[i] = gMonSecondaryCost[mon][i][1]".

ironfist/src/cpp/combat/creatures.h
```diff
@@ -110,6 +110,8 @@ enum CREATURE_FLAGS : __int16
 struct tag_monsterInfo
 {
   __int16 cost;
+  //__int16 secondary_cost_id;
```
jkoppel: Woahwoahwoah -- wait, what?

I encourage you to actually play a week of the game with these changes. I expect you will get nonsense values for creature growth because the original game's code is reading the "growth" field, but it's moved. I recommend reviewing this wiki article: http://projectironfist.pbworks.com/w/page/57731943/Interacting%20with%20the%20Original%20Code

ironfist/src/xsd/creatures_xml.xsd
```diff
@@ -21,11 +21,17 @@
...
+	<xs:attribute name="resource" type="xs:string" use="required"/>
+	<xs:attribute name="cost" type="xs:int" use="required"/> 
+  </xs:complexType>
+  
   <xs:complexType name="creature_t">
     <xs:choice maxOccurs="unbounded">
       <xs:element ref="damage" minOccurs="1" maxOccurs="1"/>
       <xs:element ref="random-spawn" minOccurs="1" maxOccurs="1"/>
       <xs:element ref="creature-attribute"/>
+	  <xs:element ref="secondary-cost"/>
```
jkoppel: Fix indentation

ironfist/src/xsd/creatures_xml.xsd
```diff
@@ -21,11 +21,17 @@
     <xs:attribute name="name" type="xs:string" use="required"/>
   </xs:complexType>
   
+  <xs:complexType name="secondary-cost_t">
```
jkoppel: This is the right design. Good job!

ironfist/src/cpp/combat/creatures.h
zhxsxuan: This was my first attempt. It definitely did not work. Let me remove these two commented lines.

zhxsxuan:
@jkoppel Ready for the second round.

ironfist/src/cpp/combat/creatures.cpp
```diff
@@ -52,16 +54,9 @@ attributeNameTableEntry creatureAttributeNameTable[] = {
...
 
-secondaryCostIdxTableEntry secondaryCostIdxTable[] = {
-	{"wood", 0},
-	{"mercury", 1},
-	{"ore", 2},
-	{"sulfer", 3},
-	{"crystal", 4},
-	{"gems", 5}
-};
+string resourceName[] = { "wood", "mercury", "ore", "sulfur", "crystal", "gems", "gold" };
```
jkoppel: This is dangerous. You're no counting on the order declared here to be the same as the one given in another file. I think I already suggested a decent way of doing this in my previous comments.

zhxsxuan:
@jkoppel I removed the dangerous array for resource names.

ironfist/src/cpp/combat/creatures.h
jkoppel: A belated apology for ripping into you above erroneously. It took me a while to realize that you hadn't removed things from the structure, and figured this out on your own. Yay for testing!

ironfist/src/cpp/adventure/adv.h
```diff
@@ -51,7 +53,7 @@ enum RESOURCES {
   RESOURCE_SULFUR = 3,
   RESOURCE_CRYSTAL = 4,
   RESOURCE_GEMS = 5,
-  RESOURCE_GOLD = 6,
+  RESOURCE_GOLD = 6,
```
jkoppel: This is a sign that you should merge in master ;)

ironfist/src/cpp/combat/creatures.cpp
```diff
@@ -167,4 +205,43 @@ void UnloadCreatures() {
...
+
+void __fastcall GetMonsterCost(int mon, int *const costs) {
+	
+	int i;
+	for (int i = 0; i < NUM_SECONDARY_RESOURCES; i++) {
+		costs[i] = gMonSecondaryResourceCost[mon][i];
+	}
+	costs[RESOURCE_GOLD] = gMonsterDatabase[mon].cost;
+
+	if (!customed_secondary_cost) {
```
jkoppel: I think what's left is to merge in master and then replace most of the logic in here with just putting this data into creatures.xml, and then we're good to go!

zhxsxuan:
@jkoppel I merged. But it looks like there is no changes after the merge. Also, I think it is redundant to put the logic value customed_second_cost into creatures.xml. Because the children element <secondary-cost> is optional in creatures.xml. That logic value is naturally hidden there. The value is true if there is <secondary-cost> element and false otherwise. Adding additional entry for this logic into creatures.xml may create additional errors because there maybe cases when a user set this value true but there is no <secondary-cost> element or there exists a <secondary-cost> element but the logic value is false.

zhxsxuan:
@jkoppel Fixed it.

jkoppel:
I just removed the redundant "int i" and "return" lines. This is ready. Good job!

PR #19, opened 2016-06-26
Upgrading Lua and loading script from a file instead of a string

jkoppel:
This seems to have fixed problems that @Kert  was having with scripting, and also improves our Lua error messages.

@rems75  Please review.

### DIFF
--- a/ironfist/src/cpp/game/save.cpp
+++ b/ironfist/src/cpp/game/save.cpp
@@ -10,12 +10,9 @@
 
 #include<iostream>
 #include<fstream>
-#include<string>
 #include<io.h>
 #include<fcntl.h>
 #include<sys/stat.h>
-
-using namespace std;
 
 extern int giMonthType;
 extern int giMonthTypeExtra;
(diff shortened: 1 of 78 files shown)


ironfist/src/cpp/game/save.cpp
```diff
@@ -527,18 +524,16 @@ int game::SaveGame(char *saveFile, int autosave, signed char baseGame) {
...
 	for(int i = 0; i < ELEMENTS_IN(this->heroes); i++) {
 		m.hero().push_back(WriteHeroXML(&this->heroes[i]));
 	}
 
-	if (GetScriptContents() != NULL) {
-		m.script(GetScriptContents());
-		SaveMapVariables(m);
-	}
+	m.script(GetScriptContents());
+	SaveMapVariables(m);
```
rems75: No need to check if there is a script anymore?

jkoppel: I remembered why I did this!

It's because GetScriptContents() now returns "" if there is none. Lack of a script is just an empty script.

However, I've confirmed that SaveMapVariables is now responsible for the crash.

PR #21, opened 2016-07-23
New VISIT_CAMPFIRE trigger with scripting support

Kert:


ironfist/src/cpp/adventure/adv.h
```diff
@@ -42,9 +42,9 @@ enum SECONDARY_SKILL
   SECONDARY_SKILL_ESTATES = 13,
 };
 
-#define NUM_RESOURCES 7
```
Kert: This is some weird end line format difference that I couldn't unstage, Same for things below

jkoppel:
This is a solid, clean job. The only things to change are to try to use the std::string version of ScriptSignal, and to fix an indentation issue.

We've been a bit lax about line endings. It's time to finally change that. I'm going to go set a policy.

jkoppel:
Merge in master with the .gitattributes addition; I think that should take care of things.

PR #23, opened 2016-07-27
Sn fix game resolution

Kert:


Kert:
I tried autocomplete in VS on read_pref thing before and it didn't work (because no #include)
Forgot to search for it in the whole project and invented a wheel instead :/

ironfist/src/cpp/init.cpp
```diff
@@ -10,6 +10,7 @@
 #include "resource/resources.h"
 #include "sound/sound.h"
 #include "town/town.h"
+#include "registry_prefs.h"
```
jkoppel: Nitpick: Please alphabetize the includes in accordance with the style guide ( http://projectironfist.pbworks.com/w/page/109715005/Coding%20Guidelines )

jkoppel:
Nice!

PR #24, opened 2016-08-02
Sn combat monster function

Kert:


ironfist/src/cpp/adventure/adv.cpp
```diff
@@ -137,4 +137,8 @@ void advManager::DoEvent(class mapCell *cell, int locX, int locY) {
 		ScriptSignal(SCRIPT_EVT_VISIT_CAMPFIRE, tmp.str());
 	}
 	this->DoEvent_orig(cell, locX, locY);
+}
+
+int advManager::CombatMonsterEvent(class hero *hero, int mon1, int mon1quantity, class mapCell *mapcell, int locX, int locY, int a8, int locX2, int locY2, int mon2, int mon2quantity, int mon2stacks, int mon3, int mon3quantity, int mon3stacks) {
```
jkoppel: If you just want to call the function, you don't need to do this.

Just declare the function "extern" in the .h file, and it will just work. :)

PR #32, opened 2016-10-30
Ds share vision

XT3RM1N8R:
I left a long series of notes while working on this in the Trello card for "ShareVision". I suggest looking at these briefly, but I want to note here that I was able to force the "ShareVision" function to fail in one save, at first, but I can no longer produce such a failure. I had previously thought that there might have been an issue with the way that the matrix was stored, but apparently that's stored with the save files as well. Somehow, the one time that produced a failure has somehow acquired the data representing the matrix with all "false" elements after previously having a functioning "ShareVision" and "PropagateVision" such that they would reveal the enemy tiles. After quite a bit of testing, I am no longer able to produce this effect. So, as far as I know, the "ShareVision" function now works! Now I'll just be forced to ponder the ramifications of my anomaly.

### DIFF
--- a/ironfist/ironfist/ironfist.vcxproj.user
+++ b/ironfist/ironfist/ironfist.vcxproj.user
@@ -4,4 +4,7 @@
     <LocalDebuggerWorkingDirectory>$(ProjectDir)..\bin</LocalDebuggerWorkingDirectory>
     <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
   </PropertyGroup>
+  <PropertyGroup>
+    <ShowAllFiles>true</ShowAllFiles>
+  </PropertyGroup>
 </Project>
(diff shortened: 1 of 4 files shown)


jkoppel:
Please remove your changes to the .vcxproj file

Kert:
You still have ironfist.vcxproj.user changed.

XT3RM1N8R:
I'm...actually not sure how that happened. I don't know what I would have done to change that.... I'll just use GitHub to remove the changes.

XT3RM1N8R:
I removed the changes to "ironfist.vcxproj.user". Not sure how they got there. Also, I removed more lines than what is seen in the commit. I also don't know why it isn't showing the actual lines that I removed.

XT3RM1N8R:
Now it updated to properly show removed content.

PR #39, opened 2016-11-11
Remember last played map and select it when the game is started

Kert:


src/cpp/shared/game/game.cpp
```diff
@@ -55,8 +56,32 @@ void game::RandomizeHeroPool() {
...
+	if (!strlen(gLastFilename)) { // game just started, no map was played yet
+		std::string lastPlayed;
+		if (xIsExpansionMap)
+			lastPlayed = read_pref<std::string>("Last Map expansion");
+		else
+			lastPlayed = read_pref<std::string>("Last Map");
+
+		if (lastPlayed.length() < 20) { // otherwise means no registry keys exist yet
+			strcpy(gMapName, lastPlayed.c_str());
+			strcpy(this->mapFilename, lastPlayed.c_str());
```
jkoppel: Wow. This easy? Awesome!

src/cpp/shared/game/game.cpp
```diff
@@ -55,8 +56,32 @@ void game::RandomizeHeroPool() {
...
+		}
+	}
+	this->InitNewGame_orig(a);
+}
+
 void game::NewMap(char* mapname) {
 	send_event(mapAction, mapname);
+	if (xIsExpansionMap)
+		write_pref("Last Map expansion", std::string(gMapName));
```
jkoppel: Move these strings into a named constants in prefs.h

Kert: We have some more registry related code somewhere but it's not using named constants. I didn't do this here for the sake of consistency. Probably this is not related to this branch.

jkoppel: Hmmm......why wouldn't I have demanded that? My normal exception to the "name all string constants" rule is if all uses fit on the same screen, which it looks like it does in this case. Merging now.

PR #40, opened 2016-11-11
Sn map put army

Kert:


ironfist/src/cpp/adventure/adv.cpp
```diff
@@ -137,4 +137,16 @@ void advManager::DoEvent(class mapCell *cell, int locX, int locY) {
 		ScriptSignal(SCRIPT_EVT_VISIT_CAMPFIRE, tmp.str());
 	}
 	this->DoEvent_orig(cell, locX, locY);
+}
+
+int advManager::MapPutArmy(int x, int y, int monIdx, int monQty) {
+	int cellIdx = x * gpGame->map.height + y;
+	gpGame->map.tiles[cellIdx].objectIndex = monIdx;
+	gpGame->map.tiles[cellIdx].extraInfo = monQty;
+	gpGame->map.tiles[cellIdx].objTileset = 12;
```
jkoppel: We should already have constants for this. If they're not already defined in some .h file, then you can get them from  HEROES2W.h

Kert: I tried to find something related this but no success.

jkoppel: Oh. Looks like this is actually in EDITOR2W.h . This one is OBJ_TILESET. Go ahead and put that enum somewhere.

ironfist/src/cpp/adventure/map.h
```diff
@@ -377,6 +377,61 @@ enum ADVENTURE_MAP_LOCATION {
   TILE_HAS_EVENT = 128,
 };
 
+enum OBJ_TILESET : __int8 {
```
jkoppel: :D

PR #42, opened 2016-11-23
Ds full hero movement points in castle

XT3RM1N8R:
I'm so sorry that this took so long...
It works now though! And I learned some cool things!


jkoppel:
Awesome! Great work.

Over time, I'm going to get more persnickety about code quality. Right now, there are two things I need you to change: the constants "2" for difficulty and personality need to be replaced with named constants. As part of this, you will need to create named constants for all difficulty levels and AI personality types, probably in game.h .

Bonus points if you can tell me why this is important.

XT3RM1N8R:
I'll check to see which difficulty level "2" represents and then add named constants for it in the game.h file. I somehow forgot that actual difficulty levels exist.

We name constants to improve code readability and make modifications easier.

jkoppel:
Yes. The reason why this is more important than the other constants like "50" is because, if we hypothetically want to add a new difficulty level and change all the difficulties to  different numbers, it would be *ahem* difficult without making these constants.

XT3RM1N8R:
I'll get that soon. Currently trying to figure what personality types are...

XT3RM1N8R:
I find the existence of both a "gameDifficulty" and "difficulty" variable rather frustrating and confusing

XT3RM1N8R:
Apparently, personality types are some sort of property unique to each player, somehow. Not quite sure I understand the mechanics in how it affects gameplay, but I don't think I need to at the moment. (BUT I may look into that later!)

...Actually, it seems to be as I expected: the personality type gives the player advantages and handicaps related to their personality type, for example, the player gets a movement point boost in CalcMobility if the player has an EXPLORER personality type.

I have added enumerations for the game difficulties and personality types in "game.h".
Definitions are just beautiful.

PR #43, opened 2016-11-23
Fixed the crash on delete hero

Kert:
This prevents the crash, but adding a hero later on may possible crash the game when the map is loaded again. The cause is unknown


jkoppel:
Wow, good catch!

Use the FREE macro instead, defined in base.h . It includes __FILE__ and __LINE__ for you.

PR #49, opened 2016-12-05
Ds fix full hero movement points in castle

XT3RM1N8R:
Fixed crash on first successive New Game of current running instance of program by adding exception for nullptr stored in "cell"


src/cpp/shared/adventure/hero.cpp
```diff
@@ -157,44 +157,45 @@ void hero::TakeArtifact(int art) {
...
-      if (!gbHumanPlayer[ownerIdx]) {
-         if (gpGame->difficulty >= DIFFICULTY_HARD) {
+  int points;
+  int MOVEMENT_POINTS_TERM_CREATURE_MAX = 1500;
+  int PLAYER_FIRST = 0;
+  int PLAYER_LAST = 5;
+  float gfSSLogisticsMod[] = { 1.0,  1.1,  1.2,  1.3 };
+  
+  mapCell* cell = gpAdvManager->GetCell(this->x, this->y);
+  if (cell != nullptr) {  //This is here because CalcMobility() is called on the first successive start of a New Game in the current running instance of the program
```
jkoppel: I think this will be much more readable if you just change this to:

if (cell == nullptr) {
  return this->CalcMobility_orig();
}

....

PR #51, opened 2016-12-16
Implemented a registry toggle for analytics activity monitoring

Kert:


jkoppel:
I think we're standardizing on two-space indentation.

PR #52, opened 2016-12-17
Ds bug fix scroll saved with null spell

XT3RM1N8R:
Modification of XSD file was determined to be unnecessary, as the real issue lied in a missing line in the save.cpp file that saved the scrollSpell data to the proper sub-element of the artifacts complex element in the XML tree. The XSD file could be modified to properly reflect the structure of the hero class in that the scrollSpell is a direct member of the hero class, but the current structure of the XML tree has been determined to be better designed, anyways, and modification of the XSD file would obligate further modification of the save.cpp file--or so I think. I would like to give a special thanks to Kert for discovering that the save.cpp file was not already saving the scrollSpell data to the XML tree.


jkoppel:
Good work!

src/cpp/shared/game/save.cpp
```diff
@@ -244,7 +244,8 @@ ironfist_map::hero_t WriteHeroXML(hero* hro) {
 
 	for(int i = 0; i < ELEMENTS_IN(hro->artifacts); i++) {
 		hx.artifact().push_back(ironfist_map::hero_t::artifact_type(hro->artifacts[i]));
-	}
+    hx.artifact().back().spell(hro->scrollSpell[i]); // This will save this in the sub-element of the complex element "artifact",
```
jkoppel: Fix indentation.

XT3RM1N8R:
Indentation is fixed here, but the document will need to be revised at a later time, when I remember how to use find and replace on a TAB element

PR #55, opened 2016-12-25
Changed default target folder to "GOG Games"

XT3RM1N8R:
This really just forces users to select the correct folder while keeping
the installer from making a new subfolder under the name of the default
path folder, as it realizes that it is already in that folder;
therefore, it simply chooses the selected folder as the target folder,
instead of creating a subfolder within the selected folder.

### DIFF
--- a/installer/ironfist.nsi
+++ b/installer/ironfist.nsi
@@ -1,7 +1,7 @@
 outfile "ironfist-install.exe"
 Name "Project Ironfist"
 
-installDir "C:\GOG Games\Heroes of Might and Magic 2 GOLD"
+installDir "C:\GOG Games\"
 
 Page directory
 Page instfiles


jkoppel:
So.....what happens if the user just leaves it as the default?

XT3RM1N8R:
The approach here forces the user to select a target folder. If it's left as the default, it may install in the wrong place. I have just realized that this solution also poses problems, because some users may not have a GOG Games folder by choice. I'm not sure if it's better to try to guess the folders that they have or to force the user to select the proper place. For the time being, I'm forcing the users to select the proper place.

jkoppel:
I don't really see this as fixing the problem, because users already don't realize that they need to set the folder to be the same as their GOG installation directory.

XT3RM1N8R:
Hmm...so would it be more desirable to try to find the GOG installation directory or would it be more desirable to simply inform users that they must set the folder to be the same as the GOG installation directory?

PR #56, opened 2016-12-26
Ds installer subfolder fix

XT3RM1N8R:


jkoppel:
I repeat my earlier comment.

Maybe you can try to have it fail with some message if it doesn't find Heroes 2 installed in the folder it installs into?

XT3RM1N8R:
I didn't read the previous comment, before hand; sorry

XT3RM1N8R:
I'll pursue the route that you suggested

PR #57, opened 2016-12-26
Ds installer subfolder fix

XT3RM1N8R:
I have modified the NSI file to what I think satisfies the specifications; it tells users to select a folder containing the game, tells users that the game isn't found if an improper folder is selected, and it quits the installer upon failure.

I don't know if I should keep the new installer in the folder with the new NSI or if I should only keep the new NSI.

### DIFF
--- a/installer/ironfist.nsi
+++ b/installer/ironfist.nsi
@@ -1,21 +1,40 @@
 outfile "ironfist-install.exe"
 Name "Project Ironfist"
 
-installDir "C:\"
+InstallDir "C:\"
 
 Page directory
 Page instfiles
 
-section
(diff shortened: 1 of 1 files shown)


jkoppel:
Code looks good. How did you test it?

XT3RM1N8R:
I tested it by trying to install it in all 3 potential folders. If I remember correctly, I tried both ways; the "Browse" button and by manually typing folder name in. I also made sure that the failure screen popped up successfully, but I guess that this happened when I tested the folder that didn't have the game.

jkoppel:
I'm satisfied with that answer. Good job!

PR #59, opened 2016-12-30
Ds shrine string fix

XT3RM1N8R:
DONE (?)

### DIFF
--- a/src/cpp/shared/adventure/adv.cpp
+++ b/src/cpp/shared/adventure/adv.cpp
@@ -5,6 +5,7 @@
 #include "game/game.h"
 #include "gui/dialog.h"
 #include "scripting/callback.h"
+#include "game/spells.cpp"
 #include "prefs.h"
 
 #include <sstream>
@@ -128,7 +129,7 @@ void advManager::DoEvent(class mapCell *cell, int locX, int locY) {
         gText,
         "%s'%s'.  ",
         "{Shrine of the 1st Circle}\n\nYou come across a small shrine attended by a group of novice acolytes.  In exchange for your protection, they agree to teach you a simple spell - ",
-        cell->extraInfo - 1;
+        gSpellNames[cell->extraInfo - 1];
       break;
     }
     case LOCATION_SHRINE_SECOND_ORDER: {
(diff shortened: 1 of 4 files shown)


src/cpp/shared/resource/resourceManager.h
```diff
@@ -76,6 +76,8 @@ class resourceManager : baseManager
 
 extern resourceManager* gpResourceManager;
 
+extern void __fastcall WaitEndSample(struct SAMPLE2, int);
```
jkoppel: Shouldn't this be in sound.h?

src/cpp/shared/adventure/adv.cpp
```diff
@@ -118,9 +121,53 @@ void game::MakeAllWaterVisible(int player) {
...
+  hero *hro = &gpGame->heroes[gpCurPlayer->curHeroIdx];
   int locType = cell->objType & 0x7F;
+  SAMPLE2 res2 = NULL_SAMPLE2;
   ScriptCallback("OnLocationVisit", locType, locX, locY);
-  this->DoEvent_orig(cell, locX, locY);
+
+  switch (locType) {
+    case LOCATION_SHRINE_FIRST: {
+      sprintf(gText, "{Shrine of the 1st Circle}\n\nYou come across a small shrine attended by a group of novice acolytes.  In exchange for your protection, they agree to teach you a simple spell - '%s'.  ", gSpellNames[cell->extraInfo - 1]);
+      break;
```
jkoppel: This is spaghetti. You can factor these out into a separate "HandleSpellShrine" function to make the control-flow easier to follow.

XT3RM1N8R: Ok. Can/Should I modify this branch and re-PR it into the master?

src/cpp/shared/resource/resourceManager.h
XT3RM1N8R: I'm not currently sure of this, but I will look into this further.

src/cpp/shared/adventure/adv.cpp
jkoppel: Yes.

src/cpp/shared/resource/resourceManager.h
XT3RM1N8R: Ok, so I remember that Sergei told me to put this specifically in resourceManager.h, but I know that this is a global variable and that I considered putting it in adv.h, before he told me to put it in resourceManager.h

XT3RM1N8R: I'm not actually sure about the specifics of the workings of this function--nevermind, I just figured it out, I think. It looks like it is used to factor in some game delays (or lag?) in the consideration of when to poll sounds in-game. I see why this would be in sound.h, but I was never given the reasoning as to why Kert wanted it in resourceManager.h, so I guess I'll run it by him, too

src/cpp/shared/adventure/adv.cpp
XT3RM1N8R: I'm going to just put the switch statement in a separate function, unless you would like me to change something else. I hope that is what you meant.

PR #60, opened 2017-01-10
Ds shrine string fix

XT3RM1N8R:
Moved WaitEndSample global declaration to sound.h because it addresses sound polling and moved spell shrine handling of DoEvent case for spell shrines to HandleSpellShrine function for better reading and comprehension.


src/cpp/shared/adventure/adv.cpp
```diff
@@ -122,11 +121,22 @@ void game::MakeAllWaterVisible(int player) {
 
 void advManager::DoEvent(class mapCell *cell, int locX, int locY) {
   hero *hro = &gpGame->heroes[gpCurPlayer->curHeroIdx];
-  int locType = cell->objType & 0x7F;
+  int LocationType = cell->objType & 0x7F;
```
jkoppel: The convention is that variable names should be lowercase.

src/cpp/shared/adventure/adv.cpp
```diff
@@ -122,11 +121,22 @@ void game::MakeAllWaterVisible(int player) {
...
+  this->UpdateTownLocators(1, 1);
+  this->UpdBottomView(1, 1, 1);
+  this->UpdateScreen(0, 0);
+  gpSoundManager->SwitchAmbientMusic(giTerrainToMusicTrack[this->currentTerrain]);
+  WaitEndSample(res2, res2.sample);
+  CheckEndGame(0, 0);
+}
 
-  switch (locType) {
+void advManager::HandleSpellShrine(class mapCell *cell, int LocationType, hero *hro, SAMPLE2 res2, int locX, int locY) {
```
jkoppel: This is not what I meant at all. We can chat about this.

src/cpp/shared/adventure/adv.cpp
XT3RM1N8R: I'll fix this

src/cpp/shared/adventure/adv.cpp
XT3RM1N8R: I made sure to move the logic for determining whether or not a location was a shrine or not (thus, determining whether or not to handle a spell shrine event) to the main modified DoEvent function, before the potential call of HandleSpellShrine. Sorry for the confusion

jkoppel:
You've solved the biggest issue in this code. I still have many problems with it, but it's ready for merging.

XT3RM1N8R:
What other issues were there?

PR #62, opened 2017-01-24
Sn indigo panther

Kert:


src/cpp/shared/combat/animation.cpp
```diff
@@ -1,5 +1,152 @@
+#include <string.h>
 #include "combat/animation.h"
 
+void BuildTeleporterTempWalkSeq(SMonFrameInfo *frm, int last, int notFirst, bool closeMove) {
```
jkoppel: It's nice that these are in a separate file to make them easier to follow. As a first candidate for refactoring, I think all of this can be done declaratively -- you can write down some data that dictates how these animations will be created, and it should be easier to read.

src/cpp/shared/combat/animation.cpp
```diff
@@ -1,149 +1,79 @@
...
+          frm->animationLengths[animType]);
+  memcpy(&frm->moveAnimInfoIgnored[frm->animationLengths[ANIMATION_TYPE_WALKING]],
+          animInfo,
+          frm->animationLengths[animType]);
+  frm->animationLengths[ANIMATION_TYPE_WALKING] += frm->animationLengths[animType];
+}
+
+static void AllocAnimation2(SMonFrameInfo *frm, int animType, char *animInfo) {
+  memcpy(&frm->animationFrameToImgIdx[ANIMATION_TYPE_WALKING][frm->animationLengths[ANIMATION_TYPE_WALKING]],
+          frm->animationFrameToImgIdx,
```
jkoppel: It looks like this is the only difference between  AllocAnimation and AllocAnimation2, and this looks like a type error.

jkoppel: I think this is equivalent to AllocAnimation with animType == 0, i.e.: animation ANIMATION_TYPE_START_MOVE .

src/cpp/shared/combat/army.cpp
```diff
@@ -0,0 +1,1626 @@
...
+    this->animationFrame = 0;
+    gpCombatManager->DrawFrame(1, 1, 0, 0, 75, 1, 1);
+  }
+}
+
+// ironfist function
+bool army::IsCloseMove(int toHexIdx) {
+  for (int j = 0; j < 6; j++) {
+    if (this->creature.creature_flags & TWO_HEXER) {
+      if (gpCombatManager->hexNeighbors[this->occupiedHex + 1][j] == toHexIdx)
```
Kert: char hexNeighbors[117][6];

src/cpp/shared/combat/army.cpp
```diff
@@ -0,0 +1,1626 @@
...
+  if (!(a->creature.creature_flags & TWO_HEXER))
+    return;
+
+  if (a->facingRight == 1)
+    a->occupiedHex--;
+  else
+    a->occupiedHex++;
+}
+
+int __fastcall OppositeDirection(signed int hex) {
```
Kert: I just ported it to C++ beacuse why not? It was a short one

src/cpp/shared/combat/army.cpp
```diff
@@ -0,0 +1,1626 @@
...
+    result = 7;
+  } else if (hex > 6) {
+    result = 6;
+  } else {
+    result = (hex + 3) % 6;
+  }
+  return result;
+}
+
+void DoAttackBattleMessage(army *attacker, army *target, int creaturesKilled, int damageDone) {
```
Kert: Would the code be shorter? Afaik we can simulate sprintf format function only with std string streams. And those aren't very beatiful

PR #64, opened 2017-03-07
Sn save xml

Kert:


src/xsd/map_xml.xsd
```diff
@@ -106,20 +106,402 @@
...
+      </xs:element>
+	  <xs:element name="field_44D" type="xs:int"/>
+	  <xs:element name="field_451" type="xs:int"/>
+	  <xs:element name="playerHandicap" minOccurs="6" maxOccurs="unbounded">
+        <xs:complexType>
+          <xs:attribute name="id" type="xs:int" use="required"/>
+          <xs:attribute name="handicap" type="xs:int"/>
+        </xs:complexType>
+      </xs:element>
+	  <xs:element name="field_459" minOccurs="6" maxOccurs="unbounded">
```
jkoppel: Watch the indentation.

src/cpp/shared/game/save.cpp
```diff
@@ -42,6 +42,18 @@ extern void __fastcall GenerateStandardFileName(char*, char*);
...
 
+template<typename T>
+void ReadArrayFromXML(T &dest, xsd::cxx::tree::sequence<ironfist_save::arrayInt_t> &src) {
+  for (int i = 0; i < src.size(); i++)
+    dest[i] = src.at(i).value();
+}
+
+template<typename T>
+void WriteArrayToXML(xsd::cxx::tree::sequence<ironfist_save::arrayInt_t> &dest, T &src) {
+  for (int i = 0; i < ELEMENTS_IN(src); i++)
```
jkoppel: ....I have trouble believing that this ELEMENTS_IN call works.

Kert: Well you can save the game and see that it works in the new savefile though :)

jkoppel: The behavior I expect to see is that ELEMENTS_IN(src) will always evaluate to 1.

Can you explain why this won't be the case?

Even if it works, you still have to explain it.

jkoppel: #include<stdio.h>

#define ELEMENTS_IN(x) ((sizeof(x))/(sizeof((x)[0])))


int foo[42];

template<typename T>
void doThing(T &src) {
  printf("Size is %d\n", ELEMENTS_IN(src));
}


int main() {
  doThing(foo);
}


-----------------

Output: 42

Mind blown. I am so confused.

jkoppel: In my example, it works because doThing got specialized to int[42]

PR #67, opened 2017-04-05
Sn unpack assets

Kert:
All our creature/spell graphics and sounds are now stored on the repo in an unpacked way.
This will let us be able to track exact file changes and it will be easier to avoid merge conflicts.
The packing script and installer making scripts were updated for this too.


jkoppel:
Boss!

Where's the Delphi code for the updated packers?

Kert:
Is it really that needed? It's one line change in each packer

jkoppel:
Yes. The source code we have should correspond to the source code we run.

PR #71, opened 2017-06-02
Sn cpp2lua classbinding

Kert:


src/cpp/shared/scripting/scripting.cpp
```diff
@@ -47,10 +47,10 @@ void MakeLuaHeroTable(lua_State *L, void *ptrAddr) {
   lua_setmetatable(L,-2);
 }
 
-void* GetPointerFromLuaClassTable(lua_State *L, int argNum) {
+void* GetPointerFromLuaClassTable(lua_State *L, int argNumber, int numArgs) {
   lua_pushstring(L, "ptr");
-  lua_gettable(L, -2);
-  void* ret = (void*)(int)lua_tonumber(L, -argNum);
+  lua_gettable(L, -numArgs-1);
+  void* ret = (void*)(int)lua_tonumber(L, -argNumber);
```
jkoppel: lua_gettable pushes something into the middle of the stack?

jkoppel:
This is excellent.

Please add an error message if the player tries to access an unsupported field of a deep-bound C++ object. 

Good work.

src/cpp/shared/scripting/scripting.cpp
Kert: As far as I understood it pushes the table with index "-numArgs-1" to the stack

jkoppel: Hmmm. It seems to me that you want lua_tonumber to run on the value which had been returned form gettable, which lies on top of the stack, and so  this only works if -argNumber == -1, i.e.: if argNumber == 1. 

And, indeed, at every caller, you do have argNumber = 1.  So why have this as a parameter?

maps/MODULES/binding.lua
```diff
@@ -74,6 +84,8 @@ battleStack_mt = {
 			return GetStackQuantity(t)
 		elseif k == "hex" then
 			return GetStackHex(t)
+		else
+			MessageBox("This field is not supported")
```
jkoppel: Good work! Still a couple of comments left to address.

jkoppel:
Yep, I'm pushing on this:

Instead of having separate argNumber and numArgs arguments, pass in the position of that table on the stack.

Once that's changed, this will be good to go.

Kert:
I disagree. An example:
Suppose t - is a table passed to c++ function and x,y - regular variables (int / string / doesn't matter), f - our c++ function
I tested the cases below and here's what I got:
For f(t) the stack index needed is -2
For f(x, c) the stack index needed is also -2
For f(c, x) the stack index needed is -3
For f(c, x, y) the stack index needed is -4
I don't see how a stack index passed to a function is easier to understand or maintain than 2 arguments: one of which is the number of all arguments and the other one is the position of the t.

jkoppel:
It depends on your perspective.

If you consider the caller + GetPointerFromLuaClassTable together, then it's easier to think of the argument count and position.

If you read GetPointerFromLuaClassTable as a standalone, then the argument count and position don't even make sense because you don't know what they're referring to.

One option would be to have a function that computes the stack index from the argNum/argCount, so you might have code like:

    hero* hro = (hero*)GetPointerFromLuaClassTable(L, StackIndexOfArg(1, 3));

This both lets GetPointerFromLuaClassTable be more self-contained, and still lets you think in terms of those two arguments.

PR #78, opened 2017-07-19
Update LastStand.MX2.lua to use AdvMessageBox

cometfractals:
Replaced some MessageBox's with AdvMessageBox, (and made a minor text string change)


jkoppel:
Use DIALOG_OKAY instead of the magic number 1, put spaces after commas, and this will be good to go.

PR #81, opened 2017-07-23
Ds expand lua deepbinding

XT3RM1N8R:
playerData:
color
daysLeftWithoutCastle

town:
x
y
visitingHeroIdx
visitingHero

hero:
tempMoraleBonuses
tempLuckBonuses
mobility
remainingMobility
x
y

army:
creatureIdx
hex
initialQuantity
quantity
morale
luck

### DIFF
--- a/maps/MODULES/binding.lua
+++ b/maps/MODULES/binding.lua
@@ -46,6 +46,8 @@ town_mt = {
 			return GetTownOwner(t)
 		elseif k == "faction" then
 			return GetTownFaction(t)
+		elseif k == "x" then
+			return GetTownX(t)
 		else
 			MessageBox("This field is not supported")
 		end
--- a/maps/MODULES/binding.lua
+++ b/maps/MODULES/binding.lua
@@ -12,6 +12,10 @@ hero_mt = {
 			SetHeroTempMoraleBonuses(table, value)
 		elseif key == "tempLuckBonuses" then
 			SetHeroTempLuckBonuses(table, value)
+		elseif key == "mobility" then
+			SetHeroMobility(table, value)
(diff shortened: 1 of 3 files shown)


maps/MODULES/binding.lua
```diff
@@ -46,18 +66,39 @@ town_mt = {
 			return GetTownOwner(t)
 		elseif k == "faction" then
 			return GetTownFaction(t)
+		elseif k == "x" then
+			return GetTownX(t)
+		elseif k == "y" then
+			return GetTownY(t)
+		elseif k == "visitingHeroIdx" then
```
jkoppel: I'm guessing this is one of the ones you did while sleep deprived. One thing that happens on all-nighters is it's easy to just sit down and do things without without asking why or whether you need to.

Remove this one. As I've said, indices like this should not be exposed to the scripter.

maps/MODULES/binding.lua
```diff
@@ -80,10 +127,18 @@ battleStack_mt = {
 			return GetStackSide(t)
 		elseif k == "type" then
 			return GetStackType(t)
+		elseif k == "creatureIdx" then
```
jkoppel: Rename to something like "creatureType" .

maps/MODULES/binding.lua
XT3RM1N8R: Oh, I think I forgot that you said this. My bad. Had you not reminded me here, I don't think I would have remembered. xD

maps/MODULES/binding.lua
XT3RM1N8R: I used creatureIdx because this was the actual name of the property, but I will rename it as "creatureType"

PR #84, opened 2017-08-04
Create Lua Table before passing ScriptCallback args

arpit-kalla:


src/cpp/shared/combat/army.cpp
```diff
@@ -1790,4 +1791,9 @@ void army::InitClean() {
   this->mirrorIdx = -1;
   this->armyIdx = -1;
   this->previousQuantity = -1;
-}
\ No newline at end of file
+}
+
+void army::InitClean() {
+	this->InitClean_orig();
```
jkoppel: There's no need to redefine InitClean_orig. You can just use the version in the assembly.

src/cpp/shared/combat/army.cpp
```diff
@@ -1790,4 +1791,9 @@ void army::InitClean() {
...
+	gpCombatManager->experienceForDefeating[1] += gMonsterDatabase[this->creatureIdx].hp;
```
jkoppel: You've special-cased the player? :o

src/cpp/shared/scripting/scripting.h
```diff
@@ -5,6 +5,12 @@
 #include<map>
 #include<string>
 
+extern "C" {
+#include "lua/src/lua.h"
+#include "lua/src/lualib.h"
```
jkoppel: Only the first include is needed.

And really, it's not needed. You can just write "struct lua_State;", and it will do all you need. No need to make someone have to think about why you're importing these things.

jkoppel:
I think we have to call this a loss. It's been 2.5 years. At this point, I barely remember what this is for, and it would be tough to merge.

PR #88, opened 2017-08-07
getGuildSpell

cometfractals:


src/cpp/shared/scripting/scripting.cpp
```diff
@@ -796,6 +796,18 @@ int l_getTownIDFromPos(lua_State *L) {
   return 1;
 }
 
+int l_getguildspell(lua_State *L) {	
+	town* twn = (town*)GetPointerFromLuaClassTable(L, StackIndexOfArg(1, 3));
+	int l = (int)luaL_checknumber(L, 2);
+	int n = (int)luaL_checknumber(L, 3);
+	if (twn->mageGuildSpells[l][n] != NULL)
+		{
```
jkoppel: Please read the Ironfist style notes about how to format your braces

src/cpp/shared/scripting/scripting.cpp
```diff
@@ -796,6 +796,18 @@ int l_getTownIDFromPos(lua_State *L) {
   return 1;
 }
 
+int l_getguildspell(lua_State *L) {	
+	town* twn = (town*)GetPointerFromLuaClassTable(L, StackIndexOfArg(1, 3));
+	int l = (int)luaL_checknumber(L, 2);
+	int n = (int)luaL_checknumber(L, 3);
+	if (twn->mageGuildSpells[l][n] != NULL)
```
jkoppel: You realize mageGuildSpells[l][n] is not a pointer, right? This NULL check is nonsensical.

cometfractals: I have assumed all along that mageGuildSpells[][] is a 2 dimensional array. The idea of line was to check that there is value (spell)  in the requested array position before retrieving it so the game would run smoother.

cometfractals:
A replacement PR for the same code will submitted

PR #89, opened 2017-08-07
getGuildSpellReg

cometfractals:


src/cpp/shared/scripting/scripting.cpp
```diff
@@ -805,7 +805,8 @@ void set_lua_globals(lua_State *L) {
   lua_register(L, "GetCurrentHero", l_getcurrenthero);
   lua_register(L, "GetPlayer", l_getplayer);
   lua_register(L, "GetNumHeroes", l_getnumheroes);
-  lua_register(L, "GetHero", l_gethero);
+
+	lua_register(L, "GetHero", l_gethero);
```
jkoppel: I recommend running "git diff" before you commit to make sure you don't accidentally commit changes like this. The "git add -p" command can help you only commit changes to the part of a file that you want to, and "git checkout -p" can help you undo changes to a part of a file.

cometfractals: Sorry. I was rushing.

PR #90, opened 2017-08-10
Created expansion class to handle additions to original classes. Impl…

Kert:
…emented astral dodge using it

### DIFF
--- a/src/cpp/shared/combat/army.cpp
+++ b/src/cpp/shared/combat/army.cpp
@@ -7,7 +7,7 @@
 #include "spell/spells.h"
 #include "expansions.h"
 
-extern ironfistExtra gIronfistExp;
+extern ironfistExtra gIronfistExtra;
 
 extern char *cMonFilename[]; // it's inside creature.cpp
 extern char *cArmyFrameFileNames[]; // it's inside creature.cpp
@@ -314,7 +314,7 @@ void army::DoAttack(int isRetaliation) {
       }
     }
 
-    if(gIronfistExp.combat.stack.abilityNowAnimating[primaryTarget]) {
+    if(gIronfistExtra.combat.stack.abilityNowAnimating[primaryTarget]) {
         if(primaryTarget->creatureIdx == CREATURE_CYBER_SHADOW_ASSASSIN) { // astral dodge animations
             int dodgeAnimLen = 7;
             primaryTarget->frameInfo.animationLengths[ANIMATION_TYPE_WINCE] = dodgeAnimLen;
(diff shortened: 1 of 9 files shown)


src/cpp/shared/combat/combat.cpp
```diff
@@ -426,3 +429,22 @@ int combatManager::ValidSpellTarget(int spell, int hexIdx) {
...
+void combatManager::SetupCombat(int arg0, int arg1, hero *h1, armyGroup *a1, town *t, hero *h2, armyGroup *a2, int arg2, int arg3, int arg4) {
+    SetupCombat_orig(arg0, arg1, h1, a1, t, h2, a2, arg2, arg3, arg4);
+    ironfist_exp.combat.stack.abilityCounter.clear();
+    ironfist_exp.combat.stack.abilityNowAnimating.clear();
+}
+
+void combatManager::ResetRound() {
+    ResetRound_orig();
+    for(int i = 0; i < 2; i++) {
+        for(int j = 0; j < 21; j++) {
```
jkoppel: I think you mean "j < MAX_STACKS" 

https://github.com/jkoppel/project-ironfist/blob/master/src/cpp/shared/combat/combat.h#L115

Oh crap. I guess MAX_STACKS is defined in IDA, but not in Ironfist. Please fix.

src/cpp/shared/combat/combat.cpp
```diff
@@ -426,3 +429,22 @@ int combatManager::ValidSpellTarget(int spell, int hexIdx) {
...
+            army* ptr = &gpCombatManager->creatures[i][j];
+            if (ptr) {
```
jkoppel: Do you realize that this is always going to be true?

src/cpp/shared/expansions.h
```diff
@@ -0,0 +1,18 @@
...
+        std::map<army*, int> abilityCounter;
+        std::map<army*, bool> abilityNowAnimating;
+};
+
+class combat_expansions {
+    public:
+        stack_expansions stack;
+};
+
+class ironfist_expansions {
```
jkoppel: I think I'd prefer the name "extra" to "expansions," because "expansions" can be confused with e.g.: the game expansion Price of Loyalty.

src/cpp/shared/expansions.h
```diff
@@ -0,0 +1,18 @@
+#include <map>
+#include "combat\army.h"
+
+class stack_expansions {
```
jkoppel: Let's set a convention that these names should follow the class names they're expanding. So, make this army_extra/army_expansions

src/cpp/shared/combat/army.cpp
```diff
@@ -5,6 +5,9 @@
 #include "scripting/callback.h"
 #include "sound/sound.h"
 #include "spell/spells.h"
+#include "expansions.h"
+
+extern ironfist_expansions ironfist_exp;
```
jkoppel: I think we're keeping the original HoMM II naming convention for globals: gIronfistExp,

src/cpp/shared/combat/combat.cpp
Kert: That wouldn't be true if those are nullptr-ed after deletion, but oh well

src/cpp/shared/combat/army.cpp
```diff
@@ -7,7 +7,7 @@
 #include "spell/spells.h"
 #include "expansions.h"
 
-extern ironfist_expansions ironfist_exp;
+extern ironfistExtra gIronfistExp;
```
jkoppel: gIronfistExp --> gIronfistExtra

PR #94, opened 2017-08-21
GrantSpellScroll

cometfractals:


src/cpp/shared/scripting/scripting.cpp
```diff
@@ -820,6 +820,15 @@ int l_getguildspell(lua_State *L) {
 	return 1;
 }
 
+int l_grantspellscroll(lua_State *L) {
+	hero* hro = (hero*)GetPointerFromLuaClassTable(L, StackIndexOfArg(1, 3));	
+	int art = (int)luaL_checknumber(L, 2);
```
jkoppel: Is there ever a reason to have art be anything but ARTIFACT_SPELL_SCROLL? Why does this need to take 3 arguments?

PR #96, opened 2017-08-28
Sn berserker

Kert:


Kert:
@jkoppel Please commit proper needed creature values to creature.xml or tell me what to change.

src/cpp/shared/combat/army.cpp
```diff
@@ -1803,6 +2134,6 @@ void army::InitClean() {
   this->mirrorIdx = -1;
   this->armyIdx = -1;
   this->previousQuantity = -1;
-  if(CreatureHasAttribute(this->creatureIdx, ASTRAL_DODGE))
+  if(CreatureHasAttribute(this->creatureIdx, ASTRAL_DODGE) || CreatureHasAttribute(this->creatureIdx, JUMPER))
```
jkoppel: abilityCounter is cleared every turn for the assassin, but not for the berserker. These work differently and represent different things. They should probably use different fields.

Also, I can't comment on it, but: As long as you're touching this method, can you use NELEM(this->combatSounds) instead of 7 on line 2121?

src/cpp/shared/combat/army.cpp
```diff
@@ -1648,6 +1976,9 @@ void army::MoveTo(int hexIdx) {
 }
 
 void army::MoveAttack(int targHex, int x) {
+  gMoveAttack = false;
+  if(x == 1 || (!x && gpCombatManager->combatGrid[targHex].unitOwner != -1))
```
jkoppel: If you're using x, then give it a more descriptive variable name.

src/cpp/shared/combat/army.cpp
```diff
@@ -614,6 +650,293 @@ void army::Walk(signed int dir, int last, int notFirst) {
...
+
+  // pretending we don't see obstacles at all
+  // this does nothing for creatures that don't ignore obstacles
+  for (auto i : obstacleHexes)
+    gpCombatManager->combatGrid[i].isBlocked = 1;
+
+  return res;
+}
+
+int army::ValidPath(int hex, int flag) {
```
jkoppel: I can't find any difference between this and the original ValidPath. Why is this here?

src/cpp/shared/combat/army.cpp
```diff
@@ -314,25 +315,60 @@ void army::DoAttack(int isRetaliation) {
...
+      }
+    }
+
+    if(this->creatureIdx == CREATURE_CYBER_PLASMA_BERSERKER) {
+      if(gIronfistExtra.combat.stack.abilityNowAnimating[this]) {
+        this->frameInfo.animationLengths[ANIMATION_TYPE_MELEE_ATTACK_UPWARDS] = 
+        this->frameInfo.animationLengths[ANIMATION_TYPE_MELEE_ATTACK_FORWARDS] =
+        this->frameInfo.animationLengths[ANIMATION_TYPE_MELEE_ATTACK_DOWNWARDS] = 1;
+        this->frameInfo.animationLengths[ANIMATION_TYPE_MELEE_ATTACK_UPWARDS_RETURN] = 
+        this->frameInfo.animationLengths[ANIMATION_TYPE_MELEE_ATTACK_FORWARDS_RETURN] =
```
jkoppel: Please move all this into a data structure. Maybe one array for each new attack animation? Or maybe you can have a separate .bin file for the alternate animations, and you copy from there? Just.....make this idea take fewer than 40 lines to read.

src/cpp/shared/combat/army.cpp
```diff
@@ -314,25 +315,60 @@ void army::DoAttack(int isRetaliation) {
...
-        if(primaryTarget->creatureIdx == CREATURE_CYBER_SHADOW_ASSASSIN) { // astral dodge animations
-            int dodgeAnimLen = 7;
-            primaryTarget->frameInfo.animationLengths[ANIMATION_TYPE_WINCE] = dodgeAnimLen;
-            for (int p = 0; p < dodgeAnimLen; p++) {
-              primaryTarget->frameInfo.animationFrameToImgIdx[ANIMATION_TYPE_WINCE][p] = 34 + p;
-            }
-
-            this->PowEffect(-1, 0, -1, -1);
-
-            // revert to usual animations after the first received attack
```
jkoppel: Where did this revert code go?

src/cpp/shared/combat/army.cpp
```diff
@@ -614,6 +650,293 @@ void army::Walk(signed int dir, int last, int notFirst) {
   }
 }
 
+int army::FindPath(int knownHex, int targHex, int speed, int flying, int flag) {
+  int res;
+  std::vector<int> obstacleHexes;
+
+  if(!IsAICombatTurn() && CreatureHasAttribute(this->creatureIdx, JUMPER) && gIronfistExtra.combat.stack.abilityCounter[this]) {
+    for (int i = 0; i < 117; i++) {
```
jkoppel: Please replace all relevant uses of 117 in the cpp code with NUM_HEXES or something.

src/cpp/shared/combat/army.cpp
```diff
@@ -614,6 +650,293 @@ void army::Walk(signed int dir, int last, int notFirst) {
...
+        obstacleHexes.push_back(i);
+        gpCombatManager->combatGrid[i].isBlocked = 0;
+      }
+    }
+  }
+  res = this->FindPath_orig(this->occupiedHex, targHex, this->creature.speed, 0, flag);
+
+  // pretending we don't see obstacles at all
+  // this does nothing for creatures that don't ignore obstacles
+  for (auto i : obstacleHexes)
```
jkoppel: Are we officially using C++11 now?

src/cpp/shared/combat/army.cpp
```diff
@@ -614,6 +650,293 @@ void army::Walk(signed int dir, int last, int notFirst) {
...
 }
 
+int army::FindPath(int knownHex, int targHex, int speed, int flying, int flag) {
+  int res;
+  std::vector<int> obstacleHexes;
+
+  if(!IsAICombatTurn() && CreatureHasAttribute(this->creatureIdx, JUMPER) && gIronfistExtra.combat.stack.abilityCounter[this]) {
+    for (int i = 0; i < 117; i++) {
+      if (gpCombatManager->combatGrid[i].isBlocked != 0 && !IsCastleWall(i)) {
+        obstacleHexes.push_back(i);
```
jkoppel: Woah! I'm impressed it's this simple.

src/cpp/shared/combat/army.cpp
```diff
@@ -614,6 +650,293 @@ void army::Walk(signed int dir, int last, int notFirst) {
...
+	char _1[8];
+	PathfindingInfo mainDataStructure[1024];
+	PathfindingInfo *field_2414;
+	int field_2418;
+	int field_241C[63];
+	searchArray();
+};
+#pragma pack(pop)
+
+void army::ArcJump(int fromHex, int toHex) {
```
jkoppel: Where did this code come from? Portions of it look decompiled, and I can't read it.

Is it possible to make this share more code with ArcShot?

src/cpp/shared/combat/army.cpp
```diff
@@ -614,6 +650,293 @@ void army::Walk(signed int dir, int last, int notFirst) {
...
+      int destHex = this->GetAdjacentCellIndex(this->occupiedHex, dir);
+      if(this->creatureIdx == CREATURE_CYBER_PLASMA_BERSERKER && gIronfistExtra.combat.stack.abilityCounter[this]) {
+        if(gMoveAttack && hexIdxb < 4) { // less than 4 hexes from enemy
+          // find last hex
+          for(int h = hexIdxb; h >= 0; --h) {
+            dir = *((BYTE *)&gpSearchArray->field_2418 + h);
+            destHex = this->GetAdjacentCellIndex(this->occupiedHex, dir);
+            this->occupiedHex = destHex;
+          }
+          this->ArcJump(initialHex, destHex);
```
jkoppel: Is this the main thing that changed here?

src/cpp/shared/combat/army.cpp
Kert: It's still a counter for ability uses, so I thought it's logical to reuse the same field.
I didn't understand the NELEM thing you mentioned

src/cpp/shared/combat/army.cpp
Kert: It's something like a flag that tells that the moveattack is just move or move-and-attack. But it works differently for AI turns that's why I didn't rename it

src/cpp/shared/combat/army.cpp
Kert: The code goes to return 0 here because we don't want standing animation to show

src/cpp/shared/combat/army.cpp
Kert: It's from catapult code. They influence different things and I can't think of anything that makes it reuse the same code without making it harder to read

src/cpp/shared/combat/army.cpp
Kert: I "imported" it because -maybe- I could use it, but turned out it was not needed. I left it here because maybe it will be useful in future?

src/cpp/shared/combat/army.cpp
Kert: I don't know, any objections?

src/cpp/shared/combat/army.cpp
Kert: Tricking the mind of units~

src/cpp/shared/combat/army.cpp
jkoppel: NELEM: I meant this: https://github.com/jkoppel/project-ironfist/blob/master/src/cpp/shared/base.h#L12

src/cpp/shared/combat/army.cpp
Kert: Line 328?

src/cpp/shared/combat/army.cpp
jkoppel: How would you implement a creature that both has jump and shadow dodge?

src/cpp/shared/combat/army.cpp
jkoppel: Oy. Then comment and explain what it does.

src/cpp/shared/combat/army.cpp
jkoppel: Can you use an API like this? http://liballeg.org/stabledocs/en/alleg013.html#do_arc

src/cpp/shared/combat/army.cpp
Kert: Well then that's a more general problem since current code doesn't support more than 1 ironfist ability properly

src/cpp/shared/combat/army.cpp
Kert: Oh right! I can make an array of coordinates using a function (I guess you replied in the wrong place)

jkoppel: Doing an array also works. The suggestion is to have a callback, and invoke the callback once for each point in the arc.

Get familiar with refactoring code to use callbacks if you're not already. https://www.joelonsoftware.com/2006/08/01/can-your-programming-language-do-this/

jkoppel: Oh....yes, I did respond in the wrong place.

src/cpp/shared/combat/army.cpp
jkoppel: Okay. I approve it staying.

src/cpp/shared/combat/army.cpp
jkoppel: If everyone can compile it, then I approve.

Are you using VS 2015?

src/cpp/shared/combat/army.cpp
jkoppel: "Well then that's a more general problem since current code doesn't support more than 1 ironfist ability properly"

Wait, what? It should be trivial to give the assassins strike-and-return, right?

src/cpp/shared/combat/army.cpp
Kert: Yes

src/cpp/shared/combat/army.cpp
Kert: Strike-and-return is not using the counter. I meant those abilities that use a counter

src/cpp/shared/combat/army.cpp
Kert: I don't think it's easier to make and use and understand a function that calculates a new point from a previous point of an arc, than to calculate all the points in the arc at once and return them as an array. Not sure if I understood your idea of "callbacks" correctly..

jkoppel: No, that's not what I suggested. Look at do_arc from Allegro and understand how it's used.

src/cpp/shared/combat/army.cpp
jkoppel: Current code doesn't support more than one ability that uses the counter, because you designed it in a way so that only one ability can use the counter.

Change the design.

src/cpp/shared/combat/army.cpp
jkoppel: I re-emphasize this comment.

src/cpp/shared/combat/creatures.cpp
```diff
@@ -63,32 +63,39 @@ SecondaryResourceNameTableEntry SecondaryResourceNameTable[] = {
...
+      tmp.push_back(0);
+    }    
+    ironfistAttributeTable.push_back(tmp);
+  }
 }
 
-void GrantCreatureAttribute(int id, const char* name) {
-	for(int i = 0; i < ELEMENTS_IN(ironfistAttributeNames); i++) {
-		if(0 == strcmp(ironfistAttributeNames[i], name)) {
+void GrantCreatureAttribute(int id, std::string name) {
```
jkoppel: Thanks for changing the string type.

src/cpp/shared/combat/creatures.cpp
```diff
@@ -63,32 +63,39 @@ SecondaryResourceNameTableEntry SecondaryResourceNameTable[] = {
...
+	EarlyShutdown(&name[0], "Attempted to grant attribute that does not exist. Check creatures.xml .");
 }
 
-int CreatureHasAttribute(int id, const char* name) {
-	for(int i = 0; i < ELEMENTS_IN(ironfistAttributeNames); i++) {
-		if(0 == strcmp(ironfistAttributeNames[i], name)) {
+int CreatureHasAttribute(int id, const std::string &name) {
+  if(id == -1)
+    return 0;
+	for(int i = 0; i < (int)ironfistAttributeNames.size(); i++) {
```
jkoppel: Fix indentation.

src/cpp/shared/combat/creatures.cpp
```diff
@@ -63,32 +63,39 @@ SecondaryResourceNameTableEntry SecondaryResourceNameTable[] = {
...
+void GrantCreatureAttribute(int id, std::string name) {
+	for(int i = 0; i < (int)ironfistAttributeNames.size(); i++) {
+		if(ironfistAttributeNames[i] == name) {
 			ironfistAttributeTable[i][id] = 1;
 			return;
 		}
 	}
 
-	EarlyShutdown((char*)name, "Attempted to grant attribute that does not exist. Check creatures.xml .");
+	EarlyShutdown(&name[0], "Attempted to grant attribute that does not exist. Check creatures.xml .");
```
jkoppel: This seems unsafe. Can you just use name.c_str()?

Kert: c_str() returns const char* which is not char* that EarlyShutdown needs
https://stackoverflow.com/questions/347949/how-to-convert-a-stdstring-to-const-char-or-char

jkoppel:
Good work.

Address the comment I made about .c_str()

Remove SetJumpingAnimation  and RevertJumpingAnimation from the .h file, since they're not part of the public interface.

(Preferably also mark the functions as static, although we've not consistently done that in the past.)

PR #97, opened 2017-09-12
Generalize map initialization randomization

jkoppel:
I'm creating a PR just so I can make comments properly


src/raw_decompiled/HEROES2W.h
```diff
@@ -3000,5 +2996,44 @@ struct SCmbtHero
 };
 #pragma pack(pop)
 
+/*  172 */
+#pragma pack(push, 1)
+struct sphinxMapExtra
```
jkoppel: I suspect all these arrays in structures are examples of this: https://stackoverflow.com/a/36577185/546591

jkoppel: This is just an FYI. I was making comments and scrolling up before I realized this was an automatically generated file.

src/raw_decompiled/EDITOR2W.c
```diff
@@ -26587,7 +26587,7 @@ signed int __thiscall editManager::Main(editManager *this, int a2)
   {
     if ( gbDoRandomMap )
     {
-      editManage::GenerateRandomMap(this);
+      editManager::GenerateRandomMap(this);
```
jkoppel: Thanks for fixing this typo.

I know you didn't write this, but....I'm surprised this compiles. It should be this->GenerateRandomMap();

src/cpp/shared/town/town.cpp
```diff
@@ -14,6 +14,15 @@
 
 #include<sstream>
 
+unsigned long gTownEligibleBuildMask[] = {
+  1073266591,
```
jkoppel: These are bit masks. At the very least, they should be given in hexadecimal. Or possible in a form like ~0 & ~(1 << BUILDING_SHIPYARD) & ~(1 << BUILDING_DWELLING_3_UPGRADE), which is e.g.: the mask indicating that any building save the shipyard and the upgraded dwelling 3 may be built. (Unsure about parens.)

(If you do refactor this, you'll probably want to locally define some abbreviations.)

src/cpp/shared/game/save.cpp
```diff
@@ -516,7 +513,7 @@ ironfist_save::gamestate_t WriteGameStateXML(game* gam) {
 
   for (int i = 0; i < ELEMENTS_IN(gam->mines); i++) {
     mine *m = &gam->mines[i];
-    gs.mine().push_back(ironfist_save::gamestate_t::mine_type(m->field_0, m->owner, m->type, m->guardianType, m->guadianQty, m->x, m->y));
+    gs.mine().push_back(ironfist_save::gamestate_t::mine_type(m->field_0, m->owner, m->type, m->guardianType, m->guardianQty, m->x, m->y));
```
jkoppel: Thanks for the typo fix.

src/cpp/shared/game/game.h
```diff
@@ -221,6 +240,30 @@ extern int gbInCampaign;
...
+extern int gbInNewGameSetup;
+extern int iLastMsgNumHumanPlayers;
+extern bool gbSetupGamePosToRealGamePos[];
+extern char gcColorToSetupPos[];
+extern char giVisRange[];
+extern char xIsPlayingExpansionCampaign;
+extern int giCurTurn;
+
+extern __int16 creatureTypesAndQuantities[NUM_FACTIONS][3][3];
+extern int neutralTownCreatureTypes[NUM_FACTIONS][5];
```
jkoppel: This needs a more descriptive name. This has something to do with randomly generating the creatures guarding them, right?

src/cpp/shared/game/game.h
```diff
@@ -221,6 +240,30 @@ extern int gbInCampaign;
...
+
+extern int gbInNewGameSetup;
+extern int iLastMsgNumHumanPlayers;
+extern bool gbSetupGamePosToRealGamePos[];
+extern char gcColorToSetupPos[];
+extern char giVisRange[];
+extern char xIsPlayingExpansionCampaign;
+extern int giCurTurn;
+
+extern __int16 creatureTypesAndQuantities[NUM_FACTIONS][3][3];
```
jkoppel: This needs  a more descriptive name.

src/cpp/shared/game/game.h
```diff
@@ -221,6 +240,30 @@ extern int gbInCampaign;
...
+extern __int16 giUltimateArtifactBaseX;
+extern __int16 giUltimateArtifactBaseY;
+extern __int16 giUltimateArtifactRadius;
+
+extern char gpcBeforeGSpellLimits[];
+extern int giPlayerInitialResourcesHuman[NUM_DIFFICULTIES][NUM_RESOURCES];
+extern int giPlayerInitialResourcesAI[NUM_DIFFICULTIES][NUM_RESOURCES];
+
+extern int __fastcall FindLastToken(const char *a1, char a2);
+extern char __fastcall StrEqNoCase(int *a1, int *a2);
```
jkoppel: Ummm.....isn't this in the standard library?

src/cpp/shared/game/game.h
```diff
@@ -221,6 +240,30 @@ extern int gbInCampaign;
...
 
+extern __int16 giUltimateArtifactBaseX;
+extern __int16 giUltimateArtifactBaseY;
+extern __int16 giUltimateArtifactRadius;
+
+extern char gpcBeforeGSpellLimits[];
+extern int giPlayerInitialResourcesHuman[NUM_DIFFICULTIES][NUM_RESOURCES];
+extern int giPlayerInitialResourcesAI[NUM_DIFFICULTIES][NUM_RESOURCES];
+
+extern int __fastcall FindLastToken(const char *a1, char a2);
```
jkoppel: Ditto for this.

jkoppel: Okay....if you're declaring them extern, I guess it's okay.

src/cpp/shared/game/game.h
```diff
@@ -174,25 +179,39 @@ class game {
...
+  
 	void NextPlayer();
 	void NextPlayer_orig();
 
 	void PerDay();
 	void PerDay_orig();
 
-    void ResetIronfistGameState();
-    void ShareVision(int sourcePlayer, int destPlayer);
+  void ResetIronfistGameState();
```
jkoppel: Is the indentation right for all of these?

src/cpp/shared/game/game.h
```diff
@@ -102,11 +110,10 @@ class game {
 	char lastSaveFile[251];
 	char _12[100];
 	SMapHeader mapHeader;
-	int field_44D;
-	__int16 field_451;
+	char relatedToPlayerPosAndColor[6];
 	char playerHandicap[6];
-	char field_459[6];
-	char field_45F[6];
+	char relatedToColorOfPlayerOrFaction[6];
```
jkoppel: Somewhere there is a "#define NUM_PLAYERS 6". Use it.

jkoppel: By "somewhere," I mean "at the top of this file."

src/cpp/shared/game/game.cpp
```diff
@@ -165,4 +182,77 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
+void game::GiveTroopsToNeutralTown(int castleIdx) {
+  int randomQuantity; // [sp+18h] [bp-14h]@7
+  int quantity; // [sp+18h] [bp-14h]@15
+  signed int creatureTier; // [sp+1Ch] [bp-10h]@7
+  int creatureType; // [sp+24h] [bp-8h]@0
+  int creatureTierRandomizer; // [sp+28h] [bp-4h]@4
+
+  if ((this->castles[castleIdx].x > 0 || this->castles[castleIdx].y > 0) && this->castles[castleIdx].ownerIdx < 0) {
+    creatureTierRandomizer = Random(1, 15);
+    if (giCurTurn / 10) { // As time goes on, the odds of getting smaller stacks of stronger creatures increases, but it is still possible to get larger stacks of weaker creatures
```
jkoppel: This should be "giCurTurn / 10 > 0".

src/cpp/shared/game/game.cpp
```diff
@@ -165,4 +182,77 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
+        randomLowerBound = (randomUpperBound + randomLowerBound) / 2;
+      }
+      randomizedValue = Random(randomLowerBound, randomUpperBound) / 10;
+      pointerToArmyOfHeroIdx->quantities[creatureTier2++] = randomizedValue;
+    }
+  }
+}
+
+void game::GiveTroopsToNeutralTown(int castleIdx) {
+  int randomQuantity; // [sp+18h] [bp-14h]@7
```
jkoppel: Define these variables in a more reasonable place. Also, remove the comments about how this function is compiled.

src/cpp/shared/game/game.cpp
```diff
@@ -165,4 +182,77 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
+      randomQuantity = Random(3, 5);
+    } else if (creatureTierRandomizer <= 15) {
+      creatureTier = 40;
+      randomQuantity = Random(1, 3);
+    } else {
+      creatureTier = 50;
+      randomQuantity = 1;
+    }
+    quantity = giCurTurn / 20 + randomQuantity;
+    creatureType = neutralTownCreatureTypes[this->castles[castleIdx].factionID][creatureTier / 10];
```
jkoppel: Hmmm.....so, this function could have been generalized just by changing this array rather than by extracting this function. Oh well.

src/cpp/shared/game/game.cpp
```diff
@@ -20,6 +20,23 @@ int game::GetRandomNumTroops(int creat) {
 
 extern int gbNoCDRom;
 
+__int16 creatureTypesAndQuantities[NUM_FACTIONS][3][3] = {
```
jkoppel: Needs more descriptive name. Also, heroes can never get tier 3 creatures, so part of this data is dead (i.e.: never read).

src/cpp/shared/game/game.cpp
```diff
@@ -165,4 +182,77 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
+  int creatureType; // [sp+24h] [bp-8h]@0
+  int creatureTierRandomizer; // [sp+28h] [bp-4h]@4
+
+  if ((this->castles[castleIdx].x > 0 || this->castles[castleIdx].y > 0) && this->castles[castleIdx].ownerIdx < 0) {
+    creatureTierRandomizer = Random(1, 15);
+    if (giCurTurn / 10) { // As time goes on, the odds of getting smaller stacks of stronger creatures increases, but it is still possible to get larger stacks of weaker creatures
+      creatureTierRandomizer += Random(0, giCurTurn / 10);
+    }
+    if (creatureTierRandomizer <= 5) {
+      creatureTier = 10;
```
jkoppel: This is always divided by 10......you might as well remove the division and make this number smaller.

src/cpp/shared/game/game.cpp
```diff
@@ -165,4 +182,77 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
+  int randomizedValue2;
+  int creatureTier2 = 0; // [sp+10h] [bp-8Ch]@1
+  int randomUpperBound; // [sp+14h] [bp-88h]@9
+  signed int armySlotIdx; // [sp+18h] [bp-84h]@3
+  signed int creatureTier; // [sp+18h] [bp-84h]@6
+  int randomLowerBound; // [sp+1Ch] [bp-80h]@9
+                        //__int16 creatureTypesAndQuantities[54]; // [sp+20h] [bp-7Ch]@1
+  armyGroup *pointerToArmyOfHeroIdx = &gpGame->heroes[heroIdx].army; // [sp+8Ch] [bp-10h]@1
+  bool hasTier[3]; // [sp+90h] [bp-Ch]@1
+  int TIER_ONE = 0;
```
jkoppel: const

src/cpp/shared/game/game.cpp
```diff
@@ -165,4 +182,77 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
 	if (gpGame->allowAIArmySharing) {
 		RedistributeTroops_orig(army1, army2, a1, a2, a3, a4, a5);
 	}
+}
+
+void game::SetRandomHeroArmies(int heroIdx, int isAI) {
+  int randomizedValue; // eax@1
```
jkoppel: Define these in a more opportune place.

src/cpp/shared/game/game.cpp
```diff
@@ -165,4 +182,77 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
+  int randomizedValue2;
+  int creatureTier2 = 0; // [sp+10h] [bp-8Ch]@1
+  int randomUpperBound; // [sp+14h] [bp-88h]@9
+  signed int armySlotIdx; // [sp+18h] [bp-84h]@3
+  signed int creatureTier; // [sp+18h] [bp-84h]@6
+  int randomLowerBound; // [sp+1Ch] [bp-80h]@9
+                        //__int16 creatureTypesAndQuantities[54]; // [sp+20h] [bp-7Ch]@1
+  armyGroup *pointerToArmyOfHeroIdx = &gpGame->heroes[heroIdx].army; // [sp+8Ch] [bp-10h]@1
```
jkoppel: Just call this "heroArmy" or something.

src/cpp/shared/game/game.cpp
```diff
@@ -165,4 +182,77 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
 	if (gpGame->allowAIArmySharing) {
 		RedistributeTroops_orig(army1, army2, a1, a2, a3, a4, a5);
 	}
+}
+
+void game::SetRandomHeroArmies(int heroIdx, int isAI) {
```
jkoppel: Ignore the other comments. This entire function can be greatly simplified in structure.

It should look something like:

ClearHeroArmy();

{
  // Tier 1 creatures
  int creatureType = ..., qty = ... ;
  GrantArmy(hero.army, creatureType, qty);
}

if (randomThing) {
  // Tier 2 creatures
  int creatureType = ..., qty = ... ;
  GrantArmy(hero.army, creatureType, qty);
}


Does this make sense?

src/cpp/shared/adventure/hero.cpp
```diff
@@ -14,6 +14,15 @@ char cHeroTypeInitial[13] ={'k', 'b', 's', 'w', 'z', 'n',
 	                        '\0','\0','\0','\0','\0','\0',
                             'c'};
 
+signed char gHeroSkillBonus[NUM_FACTIONS][2][4] = {
```
jkoppel: These are numbers, not characters.

src/cpp/shared/game/game.h
XT3RM1N8R: Those inconsistencies are not reflected in my version of the file; to me, the indentation looks fine.

src/cpp/shared/game/game.h
XT3RM1N8R: What?

src/cpp/shared/game/game.h
XT3RM1N8R: What?

src/raw_decompiled/EDITOR2W.c
XT3RM1N8R: Isn't that an automatically generated file?

src/cpp/shared/game/game.h
XT3RM1N8R: I'm assuming that you are referring to creatureTypesAndQuantities. If so, I'll think of something better. If you have suggestions or preferences for what you want in the name, let me know.

src/cpp/shared/game/game.h
XT3RM1N8R: Yes. I'm not sure what information I should give in the name that isn't there.

src/cpp/shared/game/game.cpp
XT3RM1N8R: Ok. Are you suggesting that this data be left out? I understand that the additional data may not be used, but I didn't feel comfortable getting rid of it.

jkoppel: You're not getting rid of it. It's still in the original decompilation.

src/cpp/shared/game/game.h
jkoppel: It's probably a tabs vs. spaces issue. Prefer spaces.

src/cpp/shared/game/game.h
jkoppel: Yes it is.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stricmp-wcsicmp-mbsicmp-stricmp-l-wcsicmp-l-mbsicmp-l

src/cpp/shared/game/game.h
jkoppel: I mean, if these were already in the original code, and you're leaving them in and not changing anything, then there's not much to be gained by replacing them with a standard library version.

src/raw_decompiled/EDITOR2W.c
jkoppel: Oh, I didn't realize this was EDITOR2W.c because I was reading bottom-up. So, this doesn't compile.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I'm just leaving this here so I can tell my self that this comment has been addressed; I understand that the comment is obsolete now.

src/cpp/shared/adventure/hero.cpp
XT3RM1N8R: Changed to signed __int8

src/cpp/shared/adventure/hero.cpp
```diff
@@ -14,7 +14,7 @@ char cHeroTypeInitial[13] ={'k', 'b', 's', 'w', 'z', 'n',
 	                        '\0','\0','\0','\0','\0','\0',
                             'c'};
 
-signed char gHeroSkillBonus[NUM_FACTIONS][2][4] = {
+signed __int8 gHeroSkillBonus[NUM_FACTIONS][2][4] = {
```
jkoppel: __int8 x = '#';

Can you tell me what's wrong with this?

You can change this array to __int8 in IDA; then the decompiler should generate this array for you with numbers instead of characters.

src/cpp/shared/game/game.cpp
XT3RM1N8R: ...interestingly enough, this is what it decompiled as, which may be concerning. Also, I'm not sure what changing it to "giCurTurn / 10 > 0" would accomplish. I don't quite understand it.

src/cpp/shared/game/game.cpp
XT3RM1N8R: It won't be at all formatted properly in the original decompilation, but I'll go ahead and omit the tier 3 data.

src/cpp/shared/game/game.cpp
XT3RM1N8R: Right. Done.

src/cpp/shared/game/game.cpp
XT3RM1N8R: A more reasonable place? I'm not sure that I understand. The comments have been removed, though.

src/cpp/shared/game/game.cpp
jkoppel: Yeah, compilers optimize "if (x != 0)" to "if (x)".  They're equivalent.

giCurTurn / 10 > 0 is the same thing as what's currently there,  but more clear.

src/cpp/shared/adventure/hero.cpp
XT3RM1N8R: It didn't occur to me that I should change the data to integer values instead of char values. I think this works anyways, but I'll go ahead and change the data to integer values instead, since it would look better.

src/cpp/shared/game/game.cpp
XT3RM1N8R: Oops.

src/cpp/shared/game/game.cpp
XT3RM1N8R: Alright.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I'm not sure I understand what you mean here. Perhaps the comment isn't showing me exactly what line you were talking about.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I see how a ClearHeroArmy function would be beneficial here and I suppose we probably should just use GrantArmy, since it's supposed to do the same thing as what happens at the end. I noticed that you suggested dividing the logic into to groups based on Tier level. This would require duplicating the most of the logic of the function for both the Tier 1 and Tier 2 groups. It would really only get rid of the FOR loop that iterates from Tier 1 to Tier 2. Is this still what you would suggest??

src/cpp/shared/game/game.h
XT3RM1N8R: I'm assuming that the standard library contains functions similar to these (I don't yet know what these functions do) and you are suggesting that I use the standard library functions instead.

src/cpp/shared/game/game.h
XT3RM1N8R: After further reading this comment, I see now that it was a development on the previous one, and that my last post on the previous comment is now obsolete.

src/cpp/shared/game/game.h
XT3RM1N8R: My previous post is now obsolete and should be ignored.

src/cpp/shared/game/game.h
XT3RM1N8R: I'm pretty sure that I got the tabs vs. spaces thing fixed, but I'll check it again.

XT3RM1N8R: The configuration looks fine for me.

src/cpp/shared/game/game.h
XT3RM1N8R: Got it. Fixed that bit.

src/cpp/shared/town/town.cpp
XT3RM1N8R: I'm not sure how you were able to tell what the bit masks were composed of, but I'll change it to Hexadecimal for now.

src/raw_decompiled/HEROES2W.h
XT3RM1N8R: Ok. The flexible member thing is pretty cool, but I don't think I see how it can be reflected here. Perhaps I don't understand it well enough.

src/cpp/shared/game/game.cpp
XT3RM1N8R: Ah, ok.

src/cpp/shared/adventure/hero.cpp
XT3RM1N8R: It was already defined as an __int8 in IDA, so I'm not sure why it decompiled as characters. I'm taking the values from IDA and entering them by hand.

src/cpp/shared/game/game.cpp
jkoppel: Which is better?

Option A:

    int i;
    int j;
     <100 lines of code go here>
    for (i = 0; i < 100; i++) {
        for (j = 0; j < i; j++) {
            ....
        }
    }



Option B:

     <100 lines of code go here>
    for (int i = 0; i < 100; i++) {
        for (int j = 0; j < i; j++) {
            ....
        }
    }

Does this help answer the question?

src/cpp/shared/game/game.cpp
jkoppel: Sorry, I didn't mean to actually suggest that be a separate function. I mean, that's a fine thing to do, but I just meant ClearHeroArmy() as a placeholder for whatever code to do that.

My read of this code is that it's actually using fairly separate logic for tier 1 and tier 2 creatures, and does not really gain anything from the loop. So, separating them makes the code more clear, and reduces the amount of state visible at any point.

jkoppel: Okay, I had a second read. So, I guess it is using a lot of common logic. 

Instead, here are my suggestions for simplifying the function:

* Use a ClearArmy function. I think this already exists.
* Use a GrantArmy function, which definitely already exists under some name.
* Create a temporary variable for creatureTypesAndQuantities[gpGame->heroes[heroIdx].factionID]
* Shorten variable names, and move variable declarations closer to their use
* Inline some of the random numbers instead of saving to a variable

I also recommend reading this article: https://hackernoon.com/trust-no-one-an-introduction-to-large-codebases-for-new-engineers-100586a87a5f . Then you'll see why I think the comments are doing more harm than good.

With all these suggestions, I think you'll get this function to about half the current length.

jkoppel: Finally, reflect on these suggestions. I want you to get to the point where things like this are obvious. If you can make your code half as long while still doing the same thing, you will be loved.

src/raw_decompiled/HEROES2W.h
jkoppel: What I mean is: These arrays are actually longer than one element. The structure is of variable size. If you ever move these definitions into code, you'll want to declare them without a fixed size.

src/cpp/shared/game/game.h
jkoppel: Okay, I'll consider this fine.

src/cpp/shared/game/game.cpp
```diff
@@ -185,40 +185,37 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
-  int TIER_TWO = 1;
-  int TIER_THREE = 2;
+  int randomQuantity;
+  int randomUpperBound;
+  signed int armySlotIdx;
+  signed int creatureTier;
+  int randomLowerBound;
+  armyGroup *heroArmy = &gpGame->heroes[heroIdx].army;
+  bool hasTier[2];
+  const int TIER_ONE = 0;
```
jkoppel: These constants add little information.

jkoppel: (I don't think they were worth to add, but they're also fine to leave in.)

src/cpp/shared/game/game.cpp
```diff
@@ -185,40 +185,37 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
+  const int TIER_TWO = 1;
+  const int LOW_QUANTITY = 1;
```
jkoppel: Create a struct as follows:

struct randCreatEntry { // Pick a better name
   CREATURE_TYPE type;
   int lowQty;
   int highQty;
};

Change creatureTypesAndQuantities to be a two-dimensional array of these structures.

Then you can remove these two lines from the code, and also use shorter variable names because there's more context. Overall, your code becomes shorter.

src/cpp/shared/game/game.cpp
```diff
@@ -185,40 +185,37 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
+    heroArmy->creatureTypes[armySlotIdx] = -1;
+    heroArmy->quantities[armySlotIdx] = -1;
   }
   for (creatureTier = 0; creatureTier < 2; ++creatureTier) {
     if (hasTier[creatureTier]) {
-      pointerToArmyOfHeroIdx->creatureTypes[creatureTier2] = creatureTypesAndQuantities[gpGame->heroes[heroIdx].factionID][creatureTier][TIER_ONE];
-      randomLowerBound = 10 * creatureTypesAndQuantities[gpGame->heroes[heroIdx].factionID][creatureTier][TIER_TWO];
-      randomUpperBound = 10 * creatureTypesAndQuantities[gpGame->heroes[heroIdx].factionID][creatureTier][TIER_THREE] + 9;
-      if (isAI) {                              //  If isAI, randomLowerBound is assigned the average of the bounds and results in the probability of higher random values
+      heroArmy->creatureTypes[creatureTier] = creatureTypesAndQuantities[gpGame->heroes[heroIdx].factionID][creatureTier][TIER_ONE];
```
jkoppel: The usage of the TIER_ONE constant doesn't really make sense on this line. Do you see why?

jkoppel: Using the randCreatureEntry struct like I suggested solves this problem as well.

src/cpp/shared/game/game.h
jkoppel: "Creature types and quantities?" So...every hero's army contains creature types and quantities. So would a list of all the creeps on the map. This is kinda like naming a variable holding the balance of my bank and investment accounts "currency types and amounts."

jkoppel: How about "randHeroArmyBounds"?

src/raw_decompiled/HEROES2W.c
```diff
@@ -29331,8 +29331,7 @@ void __thiscall game::SetupTowns(game *this)
 // 429E30: using guessed type int numSpellsOfLevel[5];
 
 //----- (0042A730) --------------------------------------------------------
-void *__thiscall game::ProcessOnMapHeroes(game *this)
-{
+void* game::ProcessOnMapHeroes() {
```
jkoppel: This is a void fuction, not a void*. Change it in the decompiler and it will be simplified.

How can you tell?

Its return value doesn't really make sense. Its called only once, and its return value is ignored.

But, if you really want to be sure, check the assembly. For all the functions that actually return a value, you'll see it deliberately move a value into eax before the function exits. If it doesn't, it's probably a void function.

jkoppel: Good work! I'll have more code quality comments when this is ready.

src/raw_decompiled/HEROES2W.h
XT3RM1N8R: Ah, ok.

src/cpp/shared/adventure/map.h
```diff
@@ -56,6 +56,41 @@ struct mapCellExtra {
...
+  char owner;
+  char couldBeHasArmy;
+  armyGroup army;
+  char field_11;
+  char heroID;
+  char artifacts[3];
+  char field_16;
+  int experience;
+  char couldBeHasSecondarySkills;
+  char secondarySkills[8];
```
jkoppel: I'm going through your commits in order; I don't know if you already changed this.

Anyway, for constants like this that are meaningful to the gameplay and likely to show up elsewhere (i.e.: in the hero class), please use a named constant instead. I think the hero class may already have a NUM_SECONDARY_SKILLS constant.

On the other hand, that "char artifacts[3]" refers to how you can only set 3 artifacts for a hero in the map editor, which does not relate to anything other that heroes in the map editor, so it's okay to leave that one as the number 3.

Does this all make sense?

src/cpp/shared/game/game.h
```diff
@@ -256,8 +256,8 @@ extern char __fastcall StrEqNoCase(int *a1, int *a2);
 extern int gbInNewGameSetup;
 extern int iLastMsgNumHumanPlayers;
 extern bool gbSetupGamePosToRealGamePos[];
-extern char gcColorToSetupPos[];
-extern char giVisRange[];
+extern signed char gcColorToSetupPos[];
```
jkoppel: Interesting. Is this what you got from demangler.org?

src/asm/heroes2_imports.inc
```diff
@@ -148,6 +148,8 @@ IMPORT_?SetRandomHeroArmies@game@@QAEXHH@Z = 1
 ?SetRandomHeroArmies@game@@QAEXHH@Z_clone EQU ?SetRandomHeroArmies_orig@game@@QAEXHH@Z
 IMPORT_?GiveTroopsToNeutralTown@game@@QAEXH@Z = 1
 ?GiveTroopsToNeutralTown@game@@QAEXH@Z_clone EQU ?GiveTroopsToNeutralTown_orig@game@@QAEXH@Z
+IMPORT_?ProcessOnMapHeroes@game@@QAEXXZ = 1
+?ProcessOnMapHeroes@game@@QAEXXZ_clone EQU ?ProcessOnMapHeroes_orig@game@@QAEXXZ
```
jkoppel: You're not calling ProcessOnMapHeroes_orig, are you? If so, then this is not needed. Removing this line helps us distinguish between functions that we replace entirely, vs. merely wrap.

src/cpp/shared/adventure/map.h
```diff
@@ -220,7 +220,7 @@ class HeroExtra {
     }
 
   char owner;
-  char couldBeHasArmy;
+  char hasArmy;
```
jkoppel: Nice.

src/cpp/shared/game/game.cpp
```diff
@@ -268,130 +267,124 @@ void game::ProcessOnMapHeroes() {
...
+        if (!isJail) {
+          this->SetVisibility(randomHero->x, randomHero->y, randomHero->ownerIdx, giVisRange[randomHero->secondarySkillLevel[3]]);
         }
+
+        FREE(ppMapExtra[ppMapExtraHeroIdx]);
+        ppMapExtra[ppMapExtraHeroIdx] = 0;
       }
-    }
   }
+}
```
jkoppel: Uhhh.....what? Does this compile? I don't think these braces match up.

src/cpp/shared/game/game.cpp
```diff
@@ -268,130 +267,124 @@ void game::ProcessOnMapHeroes() {
...
-              mightBeHeroAlreadyExists[mapExtraHero->heroID] = 1;
-              mapExtraHero->couldBeHasFaction = 1;
-            }
-            if (isJail) {
-              mapExtraHero->owner = -1;
-            } else {
-              mapExtraHero->owner = loc->objectIndex / 7; // Uh, redundant line?
-              mapExtraHero->owner = gcColorToPlayerPos[mapExtraHero->owner];
-            }
+  memset(mightBeHeroAlreadyExists, 0, 54u);// Only 54 out of 56 bytes set to zero, for whatever reason.
```
jkoppel: 56 is divisible by 4.

My guess is that mightBeHeroAlreadyExists is actually an array of length 54, but the compiler enlarged it to 56 to make these be 4-byte aligned.

Have you learned about how compilers pad things to be multiples of 4?

src/cpp/shared/game/game.cpp
```diff
@@ -268,130 +267,124 @@ void game::ProcessOnMapHeroes() {
...
+  for (coordYForRandomHero = 0; coordYForRandomHero < MAP_HEIGHT; ++coordYForRandomHero) {
```
jkoppel: In general, variable names in the reverse-engineered code may be longer than you'd actually write in code, because (a) you never have to write them, and (b) we're still figuring out what the code does when you name them, and (c) in decompiled code, they always have scope to the full function.

Please make these names shorter (maybe just "y"), and restrict the scope to the body of the for loop. e.g.: "for (int y = 0; y < MAP_HEIGHT; y++)" .

src/cpp/shared/game/game.cpp
```diff
@@ -268,130 +267,124 @@ void game::ProcessOnMapHeroes() {
...
+    for (coordXForRandomHero = 0; coordXForRandomHero < MAP_WIDTH; ++coordXForRandomHero) {
+      loc = &this->map.tiles[(coordYForRandomHero * this->map.width) + coordXForRandomHero];
+      if ((loc->objType & 0x7F) == 55 || loc->objType == 251) {
```
jkoppel: Have we talked about how this field works, where the bottom 7 bits represent the location type, and the top bit represents if this is a location with an event? It would be nice to replace this one field with two fields (of size 7 bits and 1 bit), except they sometimes write this field as if it's one field. That may just be a compiler optimization though.

We should probably replace the "& 0x7F" with a GET_LOCATION_TYPE() macro. You can then replace the 55 and 251 with constants from the ADVENTURE_MAP_LOCATION (or whatever it's called) enum.

src/cpp/shared/game/game.cpp
```diff
@@ -268,130 +267,124 @@ void game::ProcessOnMapHeroes() {
...
+          mapExtraHero->couldBeHasFaction = 1;
+        }
+
+        if (isJail) {
+          mapExtraHero->owner = -1;
+          faction = mapExtraHero->factionID;
+        } else {
+          mapExtraHero->owner = gcColorToPlayerPos[loc->objectIndex / 7];
+          faction = loc->objectIndex % 7; // Constant here (needs to be generalized)
+          if (faction == FACTION_MULTIPLE) { // Constant here (faction related)
```
jkoppel: I think this is actually representing a random faction, not multiple factions. Yes, the game uses the constant 6 in a faction-related context to mean different things in different places. In a third context, it also represents the Captain for the Knight faction. This is something that's going to cause trouble for us.

src/cpp/shared/game/game.cpp
```diff
@@ -268,130 +267,124 @@ void game::ProcessOnMapHeroes() {
...
-              mightBeHeroAlreadyExists[randomHeroIdx] = 1;
-              this->heroes[randomHeroIdx].factionID = faction;
-              if (mapExtraHero->field_11 && mapExtraHero->heroID >= 54)
-                this->heroes[randomHeroIdx].heroID = mapExtraHero->heroID;
-              mapExtraHero->heroID = randomHeroIdx;
-            }
+        if (mapExtraHero->couldBeHasFaction) {
+          this->heroes[mapExtraHero->heroID].factionID = faction;
+        } else {
+          randomHeroIdx = this->RandomScan((signed char *)mightBeHeroAlreadyExists, 9 * faction, 9, 1000, 0);// Constant here (the game might depend on the number of heroes as it relates to the number of factions)
```
jkoppel: I think both of these 9's are something like HEROES_PER_FACTION.

src/cpp/shared/game/game.cpp
```diff
@@ -268,130 +267,124 @@ void game::ProcessOnMapHeroes() {
...
+            randomHero->skillIndex[secondarySkillIdxA] = 0;
+          }
+          for (secondarySkillIdxB = 0; secondarySkillIdxB < 8; ++secondarySkillIdxB) { //assign new secondary skills
+            if (mapExtraHero->secondarySkills[secondarySkillIdxB] != -1)
+              randomHero->GiveSS(mapExtraHero->secondarySkills[secondarySkillIdxB],  *(&mapExtraHero->firstSecondarySkillLevel + secondarySkillIdxB));
+          }
+        }
 
+        if (!isJail) {
+          this->SetVisibility(randomHero->x, randomHero->y, randomHero->ownerIdx, giVisRange[randomHero->secondarySkillLevel[3]]);
```
jkoppel: Presumably, this is SECONDARY_SKILL_SCOUTING.

src/cpp/shared/game/game.cpp
```diff
@@ -268,130 +267,124 @@ void game::ProcessOnMapHeroes() {
...
         }
+
+        FREE(ppMapExtra[ppMapExtraHeroIdx]);
+        ppMapExtra[ppMapExtraHeroIdx] = 0;
```
jkoppel: NULL, not 0.

src/cpp/shared/game/game.cpp
```diff
@@ -268,130 +267,124 @@ void game::ProcessOnMapHeroes() {
...
+          loc->objType = 0;
+        }
+
+        if (mapExtraHero->couldBeHasSecondarySkills) {
+          randomHero->numSecSkillsKnown = 0;
+          for (secondarySkillIdxA = 0; secondarySkillIdxA < 14; ++secondarySkillIdxA) { //clear secondary skills
+            randomHero->secondarySkillLevel[secondarySkillIdxA] = 0;
+            randomHero->skillIndex[secondarySkillIdxA] = 0;
+          }
+          for (secondarySkillIdxB = 0; secondarySkillIdxB < 8; ++secondarySkillIdxB) { //assign new secondary skills
```
jkoppel: Another thing about decompiled code: it can't reuse variable names.

Both this and the previous source code could just be "for (int i = 0; ... ".

src/cpp/shared/game/game.cpp
XT3RM1N8R: I have not, but I'd imagine that it is related to the way that bytes are addressed and the concept of efficiency.

src/cpp/shared/game/game.cpp
XT3RM1N8R: The last change committed should have both compiled and functioned properly. I have just checked, and it does.

XT3RM1N8R: Although, I thought I had fixed the spacing issue there. It's fixed now...for now...

src/cpp/shared/game/game.cpp
XT3RM1N8R: I've been working on trying to change some of these. I suppose the x/y names would be better in a simpler form. Also, I was under the impression that all of the variables should be declared at the beginning, but for iterators like that, I'll start practicing keeping their scopes where they really matter.

src/cpp/shared/adventure/map.h
XT3RM1N8R: Indeed.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I recall discussing this somewhat. Also, I'm not entirely sure I understand that macro. I think I understand what it might do, but not how it would work.

src/cpp/shared/game/game.cpp
XT3RM1N8R: Yeah. I'm not sure how this will be addressed yet. I'm leaving this here for now, as I'm currently not working on the generalization yet.

src/cpp/shared/game/game.cpp
XT3RM1N8R: That makes sense. I'm still not confident what is so significant about the values passed into RandomScan (I think it tries to find a random that isn't covered in mightBeHeroAlreadyExists) and I'm also not confident that this will work well with the new factions.

src/cpp/shared/game/game.cpp
XT3RM1N8R: That makes sense; I'll have to look out for this.

src/cpp/shared/game/game.cpp
XT3RM1N8R: Oh, right.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I've been meaning to fix this. I'll keep these limited to the relevant scope as well.

src/cpp/shared/game/game.cpp
jkoppel: The first argument is presumably an array of heroes that have already been used.

The second and third arguments give bounds on the hero IDs that it may use.

The fourth argument is the number of random tries, presumably.
The last argument is ???.

There are a number of "special" heroes that have the next 12 or so IDs. The Cyborg heroes would need to be after that.

src/cpp/shared/game/game.cpp
jkoppel: #define GET_LOCATION_TYPE(n) ((n) & 0x7F)

src/cpp/shared/game/game.cpp
jkoppel: "I was under the impression that all of the variables should be declared at the beginning"

Where'd you get that impression?

src/cpp/shared/game/game.cpp
jkoppel: You got it.

More importantly, in x86, you can only read a dword from an address which is divisible by 4.

src/cpp/shared/game/game.h
XT3RM1N8R: Oh, these comments aren't quite in line with the the things you were trying to highlight. I thought you were talking about neutralTownCreatureTypes.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I find the use of this function to be rather strange. I have also noticed that the left bound can actually exceed the right bound in the first call to this function. Additionally I'm not sure what special heroes you are talking about and/or what you mean by "the next 12".

src/cpp/shared/game/game.cpp
XT3RM1N8R: This is just a pattern that I noticed in other functions and in the decompilations. I guess I never fully understood the convention.

src/cpp/shared/game/game.cpp
XT3RM1N8R: Oh, ok. So, I would just replace "n" with the actual ADVENTURE_MAP_LOCATION for that context?

src/cpp/shared/game/game.cpp
XT3RM1N8R: Cool.

src/cpp/shared/adventure/map.h
XT3RM1N8R: The NUM_SECONDARY_SKILLS constant doesn't exist. I'm going to define it in game.h

src/raw_decompiled/HEROES2W.c
XT3RM1N8R: This was later corrected.

src/asm/heroes2_imports.inc
jkoppel: Yeah, take a few minutes to do so.

src/cpp/shared/game/game.cpp
jkoppel: There are certain heroes such as Brother Brax, Sister Carlawna, Archibald, Roland, and Solmyr which only appear in the campaigns, or when specified in the map editor. They never spawn randomly.

src/cpp/shared/game/game.cpp
jkoppel: Nope, it's an artifact of the decompiler.

src/cpp/shared/game/game.cpp
jkoppel: You'd invoke this as GET_LOCATION_TYPE(loc->objType) .

jkoppel: Or maybe you'd define it so you can just do GET_LOCATION_TYPE(loc).

Or define loc as a class, so you can say loc->getLocationType(); 

Point is, don't spread the knowledge that the location type is stored in the bottom 7 bits all over the codebase. That will make it even more impossible to change.

src/cpp/shared/game/game.cpp
XT3RM1N8R: Ah. It sounds like I'm going to need to learn more about the heroes to generalize this properly.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I like the idea of the getLocationType() function. "loc" is of type mapCell*. This would just be adding this function to the class definition, I think.

src/cpp/shared/game/game.cpp
XT3RM1N8R: "Create a temporary variable for creatureTypesAndQuantities[gpGame->heroes[heroIdx].factionID]". So, a pointer to the element at creatureTypesAndQuantities[gpGame->heroes[heroIdx].factionID][0][0]? I'm not sure I understand.

jkoppel: Yes, except that, if you chance the elements of creatureTypesAndQuantities[x] to a structure instead of a 2D array, then it's just a pointer to the structure at creatureTypesAndQuantities[gpGame->heroes[heroIdx].factionID] . Do the structure refactoring first.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I'm not sure how to initialize these values this way unless I initialize each struct as a class with a constructor or I use a function to make the initializations. Is there something that I'm missing here?

XT3RM1N8R: Also, I don't think CREATURE_TYPE is a valid type. It looks familiar, but I'm having trouble finding it.

src/cpp/shared/adventure/adv.cpp
```diff
@@ -181,4 +181,8 @@ int advManager::MapPutArmy(int x, int y, int monIdx, int monQty) {
   gpGame->map.tiles[cellIdx].field_4_1 = 0;
   gpGame->map.tiles[cellIdx].isShadow = 0;
   return 0;
+}
+
+int mapCell::getLocationType() {
+  return this->objType & 0x7F;
```
jkoppel: :D

src/cpp/shared/game/game.cpp
```diff
@@ -216,11 +216,8 @@ void game::SetRandomHeroArmies(int heroIdx, int isAI) {
   if ((isAI < 1 ? 25 : 65) <= Random(0, 99)) { // This line is only useful if the previous line doesn't assign "true" to hasTier[TIER_TWO].
     hasTier[TIER_TWO] = 1;                     // If isAI, there is a ~35% chance that this hero will get Tier 2 creatures; if not, then a ~75% chance
   }*/
-  for (signed int armySlotIdx = 0; armySlotIdx < 5; ++armySlotIdx) {
-    heroArmy->creatureTypes[armySlotIdx] = -1;
-    heroArmy->quantities[armySlotIdx] = -1;
-  }
-  for (signed int creatureTier = 0; creatureTier < 2; ++creatureTier) {
+  this->ClearArmy(heroArmy);
```
jkoppel: ++

src/cpp/shared/game/game.cpp
```diff
@@ -233,6 +230,13 @@ void game::SetRandomHeroArmies(int heroIdx, int isAI) {
   }
 }
 
+void game::ClearArmy(class armyGroup * heroArmy) {
```
jkoppel: This actually makes more sense to put in the armyGroup class.

So, then instead of writing "this->ClearArmy(heroArmy)", you'd write heroArmy->Clear();

src/cpp/shared/game/game.cpp
```diff
@@ -382,12 +382,12 @@ void game::ProcessOnMapHeroes() {
         }
 
         if (!isJail) {
-          this->SetVisibility(randomHero->x, randomHero->y, randomHero->ownerIdx, giVisRange[randomHero->secondarySkillLevel[3]]);
+          this->SetVisibility(randomHero->x, randomHero->y, randomHero->ownerIdx, giVisRange[randomHero->secondarySkillLevel[3]]); // probably SECONDARY_SKILL_SCOUTING
```
jkoppel: I mean, replace that 3 with this: https://github.com/jkoppel/project-ironfist/blob/master/src/cpp/shared/adventure/adv.h#L30

src/cpp/shared/game/game.cpp
```diff
@@ -221,11 +221,15 @@ void game::SetRandomHeroArmies(int heroIdx, int isAI) {
     if (hasTier[creatureTier]) {
       randomLowerBound = 10 * randomHeroArmyBounds[gpGame->heroes[heroIdx].factionID][creatureTier][LOW_QUANTITY];
       randomUpperBound = 10 * randomHeroArmyBounds[gpGame->heroes[heroIdx].factionID][creatureTier][HIGH_QUANTITY] + 9;
-      if (isAI) { //  If isAI, randomLowerBound is assigned the average of the bounds and results in the probability of higher random values
+      if (isAI) { //  If isAI, randomLowerBound is assigned the average of the bounds and this results in the probability of higher values
         randomLowerBound = (randomUpperBound + randomLowerBound) / 2;
       }
-      heroArmy->creatureTypes[creatureTier] = randomHeroArmyBounds[gpGame->heroes[heroIdx].factionID][creatureTier][TIER_ONE];
-      heroArmy->quantities[creatureTier] = Random(randomLowerBound, randomUpperBound) / 10;
+      /*heroArmy->creatureTypes[creatureTier] = randomHeroArmyBounds[gpGame->heroes[heroIdx].factionID][creatureTier][TIER_ONE];
```
jkoppel: I just checked game::GiveArmy. This change works. You can remove the commented-out code.

src/cpp/shared/game/game.cpp
```diff
@@ -193,9 +193,26 @@ void game::SetRandomHeroArmies(int heroIdx, int isAI) {
...
+  }
+  if (Random(0, 99) < tierChance) {
+    hasTier[TIER_TWO] = 1;
+  } else {
+    if (isAI) {
+      tierChance = 35;
+    } else {
+      tierChance = 75;
+    }
+    hasTier[TIER_TWO] = Random(0, 99) < tierChance;
```
jkoppel: Want to know something awful?

You're correct to put 7 on line 204, but the original code actually had <= here.

src/cpp/shared/game/game.cpp
jkoppel: Yes, I probably meant CREATURES. I didn't check the name before writing it.

https://en.wikipedia.org/wiki/Struct_(C_programming_language)#Struct_initialization

XT3RM1N8R: I'm using CREATURES here.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I'm not entirely sure why it wasn't working last time--I'll acknowledge that there were more unnecessary complexities to the structure definition then--but the suggested solution is up and working.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I didn't see your last comment until I posted my previous one and refreshed the page.

src/cpp/shared/game/game.cpp
XT3RM1N8R: Noted. Covering this now. I was thinking about that earlier, but I guess I forgot about it.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I accidentally forgot to change it.

src/cpp/shared/game/game.cpp
XT3RM1N8R: Uh....whatever misunderstanding existed here is obsolete due to the conversation we had afterwards.

src/cpp/shared/game/game.cpp
```diff
@@ -194,26 +194,8 @@ void game::SetRandomHeroArmies(int heroIdx, int isAI) {
...
-      tierChance = 75;
-    }
-    hasTier[TIER_TWO] = Random(0, 99) < tierChance;
-  }
-  
-  /*hasTier[TIER_TWO] = Random(0, 99) < (isAI < 1 ? 50 : 80); // If isAI, there is a ~80% chance that this hero will get Tier 2 creatures; if not, then a ~50% chance
-  if ((isAI < 1 ? 25 : 65) <= Random(0, 99)) { // This line is only useful if the previous line doesn't assign "true" to hasTier[TIER_TWO].
-    hasTier[TIER_TWO] = 1;                     // If isAI, there is a ~35% chance that this hero will get Tier 2 creatures; if not, then a ~75% chance
-  }*/
+  hasTier[TIER_TWO] = Random(0, 99) < 87;
```
jkoppel: Gorgeous.

src/cpp/shared/game/game.cpp
```diff
@@ -185,31 +185,27 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
 
   hasTier[TIER_ONE] = 1;
   hasTier[TIER_TWO] = Random(0, 99) < 87;
 
   this->ClearArmy(heroArmy);
   for (int creatureTier = 0; creatureTier < 2; ++creatureTier) {
     if (hasTier[creatureTier]) {
-      randomLowerBound = 10 * randomHeroArmyBounds[gpGame->heroes[heroIdx].factionID][creatureTier].lowQuantity;
-      randomUpperBound = 10 * randomHeroArmyBounds[gpGame->heroes[heroIdx].factionID][creatureTier].highQuantity + 9;
+      randomLowerBound = creatureFaction[creatureTier].lowQuantity;
```
jkoppel: Did you already pre-multiply these constants by 10?

src/cpp/shared/game/game.cpp
```diff
@@ -301,7 +301,7 @@ void game::ProcessOnMapHeroes() {
         }
         randomHero = &this->heroes[mapExtraHero->heroID];
 
-        if (!isJail && mapExtraHero->relatedToName[19]) {
+        if (!isJail && mapExtraHero->name[19]) {
```
jkoppel: This is a memory error; name has length 13 now.

src/cpp/shared/game/game.cpp
```diff
@@ -322,7 +322,7 @@ void game::ProcessOnMapHeroes() {
             GiveArtifact(randomHero, mapExtraHero->artifacts[i], 1, -1);
           }
         }
-        if (mapExtraHero->name[5]) {
+        if (mapExtraHero->hasName) {
```
jkoppel: Ah, there you go. :)

src/raw_decompiled/HEROES2W.h
```diff
@@ -2958,8 +2958,7 @@ struct heroMapExtra
   int experience;
   char couldBeHasSecondarySkills;
   char secondarySkills[8];
-  char firstSecondarySkillLevel;
-  char secondSecondarySkillLevel;
+  char secondarySkillLevel;
```
jkoppel: Uhhh.....why did you delete that other field?

You can't change the size of this structure; doing so can prevent us from loading maps.

src/cpp/shared/adventure/hero.cpp
```diff
@@ -151,6 +151,14 @@ void hero::SetPrimarySkill(int skill, int amt) {
   this->primarySkills[skill] = amt;
 }
 
+void hero::ClearSS() {
+  for (int i = 0; i < 14; ++i) {
```
jkoppel: Magic number

src/cpp/shared/game/game.cpp
```diff
@@ -258,7 +258,7 @@ void game::ProcessOnMapHeroes() {
   HeroExtra *mapExtraHero;
   char isJail;
 
-  memset(heroExists, 0, 54u);// Only 54 out of 56 bytes set to zero, for whatever reason.
+  memset(heroExists, 0, 54u);
```
jkoppel: Magic number

src/cpp/shared/adventure/hero.cpp
XT3RM1N8R: I thought there was something off with that, but I couldn't find a matching macro for this relating to secondary skills.

src/cpp/shared/game/game.cpp
XT3RM1N8R: While I agree, I find that it doesn't take much to look gorgeous compared to what used to be there xD

src/cpp/shared/game/game.cpp
XT3RM1N8R: I'm not sure I understand what you are asking.

src/cpp/shared/game/game.cpp
XT3RM1N8R: It gets fixed later.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I played with "git add -p" a bit

src/raw_decompiled/HEROES2W.h
XT3RM1N8R: This...is out of date.

src/cpp/shared/game/game.cpp
XT3RM1N8R: I guess I got lazy in anticipation of the changes to come to that line. :/
Taking care of that now.

src/cpp/shared/game/game.cpp
jkoppel: Why'd you remove  the 10*?

XT3RM1N8R: I thought we spoke about this earlier in Skype. Everything was multiplied by 10 and then divided by 10. I changed the variables to "double", so the constants weren't really needed for rounding anymore.

PR #98, opened 2017-09-27
SetNumberOfCreatures

cometfractals:
Uses GetCurrentTown() as parameter. GetCurrentTown() only works within OnTownOpen()

### DIFF
--- a/src/cpp/shared/scripting/scripting.cpp
+++ b/src/cpp/shared/scripting/scripting.cpp
@@ -828,6 +828,13 @@ int l_grantspellscroll(lua_State *L) {
 
 }
 
+int l_setNumberOfCreatures(lua_State *L) {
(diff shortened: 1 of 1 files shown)


src/cpp/shared/scripting/scripting.cpp
```diff
@@ -828,6 +828,13 @@ int l_grantspellscroll(lua_State *L) {
 
 }
 
+int l_setNumberOfCreatures(lua_State *L) {
+	town* cstle = (town*)GetPointerFromLuaClassTable(L, StackIndexOfArg(1, 3));
+	int dwllng = (int)luaL_checknumber(L, 2);
+	int numcrtrs = (int)luaL_checknumber(L, 3);
```
jkoppel: This function doesn't do anything. Why are you sending a PR?

PR #105, opened 2018-01-23
Give proper Lua error messages when error encountered in callback

jkoppel:
Also reorganized scripting files a bit.

@Kert Please review

@tornado711 Note that you will encounter a merge conflict after this is merged.


jkoppel:
I'll merge this tomorrow if there are no comments before then

src/cpp/shared/scripting/funcs.cpp
```diff
@@ -0,0 +1,987 @@
...
+  lua_pushinteger(L, gpGame->week);
+  return 1;
+}
+
+int l_getmonth(lua_State *L) {
+  lua_pushinteger(L, gpGame->month);
+  return 1;
+}
+
+
```
Kert: Inconsistent amount of new lines in this file

src/cpp/shared/scripting/funcs.cpp
```diff
@@ -0,0 +1,987 @@
...
+#include "scripting/register.h"
+
+int StackIndexOfArg(int argNumber, int numArgs) {
+  return (numArgs - (argNumber - 1));
+}
+
+/************************************************ Deep binded objects ********************************************************/
+
+void MakeLuaHeroTable(lua_State *L, void *ptrAddr) {
```
Kert: What about making a single func with an additional string parameter? "hero_mt" / "town_mt" will be passed there

src/cpp/shared/scripting/funcs.cpp
```diff
@@ -0,0 +1,987 @@
+
```
Kert: Unneded newlines?

src/cpp/shared/scripting/lua_utils.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
```
Kert: Include guards instead of this? I believe the rest of the project uses guards

src/cpp/shared/scripting/callback.cpp
```diff
@@ -17,9 +18,4 @@ void ironfist_lua_push(std::string arg) {
 }
 
 void ironfist_lua_pushmulti() {
-};
-
-void ScriptCallback(const char * funcName) {
```
Kert: Have you tested using ScriptCallback with only funcName (without args) after removing this? I vaguely recall something wasn't working for me until I added this one

Kert: Have you tested using ScriptCallback with only funcName (without args) after removing this? I vaguely recall something wasn't working for me until I added this one

jkoppel: Yes. I remember that discussion and was surprised; I don't remember what the story was behind that. But, the code wouldn't compile if this didn't work.  OnMapStart was called successfully, and does not take any parameters.

src/cpp/shared/scripting/lua_utils.h
jkoppel: Yeah; either I'll move the entire project to use this, or I should just use include guards here.

jkoppel:
Handled concerns. Refactored funcs.cpp while I'm at it (I regret doing so; was quite time-consuming). I thoroughly checked that the same functions are being registered before and after.

PR #106, opened 2018-02-20
Optionally disable +2 growth from the well.

mkristofik:
Creating a clean pull request by combining several changes:
- change Well description to not mention +2 growth
- rename Well to Poisoned Well in Necromancer castle, for flavor
- undo +2 growth per week
- disable month of the plague
- update Well window creature descriptions with corrected growth rates
- added `/disable-well` command line option

### DIFF
--- a/src/cpp/shared/gui/msg.cpp
+++ b/src/cpp/shared/gui/msg.cpp
@@ -18,7 +18,7 @@ void GUISetIcon(heroWindow* hwnd, int f, char* p) {
 	GUIBroadcastMessage(hwnd, f, GUI_MESSAGE_SET_ICON, (void*)p);
 }
 
-void GUISetText(heroWindow* hwnd, int f, char* p) {
+void GUISetText(heroWindow* hwnd, int f, const char* p) {
 	GUIBroadcastMessage(hwnd, f, GUI_MESSAGE_SET_TEXT, (void*)p);
 }
 
--- a/src/cpp/shared/gui/msg.cpp
+++ b/src/cpp/shared/gui/msg.cpp
@@ -18,12 +18,24 @@ void GUISetIcon(heroWindow* hwnd, int f, char* p) {
 	GUIBroadcastMessage(hwnd, f, GUI_MESSAGE_SET_ICON, (void*)p);
 }
 
-void GUISetText(heroWindow* hwnd, int f, const char* p) {
+void GUISetIcon(heroWindow* hwnd, int f, std::string& p) {
+  GUISetIcon(hwnd, f, &p[0]);
+}
+
+void GUISetText(heroWindow* hwnd, int f, char* p) {
 	GUIBroadcastMessage(hwnd, f, GUI_MESSAGE_SET_TEXT, (void*)p);
+}
+
+void GUISetText(heroWindow* hwnd, int f, std::string& p) {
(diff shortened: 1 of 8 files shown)


jkoppel:
Nice work!

Let's talk about the Ironfist coding guidelines so I can get this merged in. Add me on Skype: aurakdraconian.

mkristofik:
Can we discuss any potential changes directly on here?  I haven't been able to make time the past couple of nights to chat on Skype about this.

jkoppel:
We can, but it will be harder to have the back-and-forth to really explain things. We can use text chat rather than voice chat.

mkristofik:
Improved on this a bit per our conversation.  Still need to move the command-line argument to the registry.

src/cpp/shared/gui/msg.cpp
```diff
@@ -18,7 +18,7 @@ void GUISetIcon(heroWindow* hwnd, int f, char* p) {
 	GUIBroadcastMessage(hwnd, f, GUI_MESSAGE_SET_ICON, (void*)p);
 }
 
-void GUISetText(heroWindow* hwnd, int f, char* p) {
+void GUISetText(heroWindow* hwnd, int f, const char* p) {
 	GUIBroadcastMessage(hwnd, f, GUI_MESSAGE_SET_TEXT, (void*)p);
```
mkristofik: This might be undefined behavior now.  I have to look whether going from const char * to plain void * is undefined.  It appears to work, but still would rather avoid UB if possible.

mkristofik:
All right, I think we're all set here.  Please let me know if you want any further changes.

One more idea I had was to lower the cost of the Well if it's been disabled.  500 gold for a building that doesn't do anything seems kinda high.  I was thinking of lowering it to 150, the cost of 1 Archer, because the only benefit left is 1 Archer strength in the turrets.

mkristofik:
It looks like there's more to changing the cost of a building than overriding GetBuildingCost.  The cost of the well is hard-coded in gNeutralBuildingCosts.  Several functions use that array directly, including the display of the costs in the castle building screen and the resources deducted when you buy it.  We could maybe do an xml file like you did for creatures, but that's probably best left for another change.

So again, I think I'm done here with this unless you have further changes you want me to make.

jkoppel:
Bahahahaha! The beauty of Ironfist's tech is that overriding arrays is just as easy as overriding functions. See what we did for artifacts and spells.

The only gotcha is that some functions may have been disassembled to that array by an offset, rather than by the base address.

Would you like to make this change? If so, I'll wait; otherwise, I'll merge.

mkristofik:
Thanks for the tips.  It might be a while before I get around to that, so let's merge this change as it is.

jkoppel:
Merging.

You're good. How'd you like to work on a more important part of the game?

mkristofik:
Sure, I'd be happy to contribute more.  I don't see any open issues on here - is there a project plan somewhere that I can look at?

BTW, do you guys delete branches after they've been merged?

jkoppel:
I just shared with you several links on Skype.

Let's video chat and I can fully bring you on-board the team. I normally assign a test before people join, but you've definitely passed.

Easiest way is to book me at https://calendly.com/jimmykoppel/30min , which has a handful of availabilities, although I'd like to find a time when you and I can both chat with our creative director Roy, who is in Poland.

PR #107, opened 2018-03-12
Enabling sound effects

mkristofik:
This doesn't do anything substantially different from the original code.  But perhaps compiling for a modern Windows allows it work.

### DIFF
--- a/src/cpp/shared/sound/cd_music.cpp
+++ b/src/cpp/shared/sound/cd_music.cpp
@@ -253,8 +253,7 @@ int soundManager::Open(int foo) {
   if (status == 0) {
     hdidriver = driver;
     AllocateSampleHandles();
-  }
-  else {
+  } else {
     std::string msg = "{Sound Error}\n\n";
     msg += AIL_last_error();
     H2MessageBox(msg);
--- a/src/cpp/shared/sound/cd_music.cpp
+++ b/src/cpp/shared/sound/cd_music.cpp
@@ -229,6 +229,11 @@ void soundManager::CDPlay(int track_number, signed int a3, int a4, int a5) {
 int soundManager::Open(int foo) {
   const DWORD volume = read_pref<DWORD>("Sound Volume");
   const int origResult = Open_orig(foo);
+
+  // The original code works on OS earlier than Windows 10.
+  if (hdidriver) {
+    return origResult;
+  }
   
   // The following does the same thing the original sound initialization code
   // did. But for some reason it works here when it didn't before.
--- a/src/cpp/shared/sound/cd_music.cpp
+++ b/src/cpp/shared/sound/cd_music.cpp
@@ -12,8 +12,10 @@
 
 #include "base.h"
 #include "game/game.h"
+#include "gui/dialog.h"
 #include "sound/sound.h"
 #include "manager.h"
+#include "prefs.h"
 #include <iostream>
 
 extern int gbNoSound;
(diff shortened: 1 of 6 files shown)


jkoppel:
It's quite odd that I'm not getting E-mails about these PRs when I am for the others.

Quite likely reason: you're sending them off a fork of this repo, instead of off this repo itself. I've added you as a collaborator. Hopefully that will fix things.

src/cpp/shared/sound/cd_music.cpp
```diff
@@ -100,9 +102,9 @@ void bass_play_track(int trknum) {
     return; //this track is already being played
 
   if (!useOpera && trknum >= 4 && trknum <= 9)
-    sprintf(trkbuf, "%s/sw/homm2_%02d.ogg", trkpath, trknum);
+    snprintf(trkbuf, sizeof(trkbuf), "%s/sw/homm2_%02d.ogg", trkpath, trknum);
```
jkoppel: Thanks for the small patch. As a small FYI: the reason why I haven't been super-strict about making people use snprintf over sprintf is that it's a bit of a lost cause; there's hundreds of uses throughout the code, usually with fixed-size buffers.

jkoppel:
We'll need to sit down together and figure out why you were having the issue so we can understand how this fixes it

Also, "gbNoSound" is very misnamed. (This name was taken from the original game's code; we didn't name it.) It actually turns off a lot of things. I don't remember why exactly it was necessary to temporarily turn it on (it's been over 5 years, and I was quite a different person then), but that variable does interact with a lot of different subsystems. So, it's not to be messed with lightly.

Similarly, the "PollSound()" function is actually a poor-man's version of non-preemptive concurrency; it should really be thought of as yielding to another thread, which does basic sound/graphics updates.

mkristofik:
`gbNoSound` is controlled by a command-line switch.  I noticed there are numerous if-statements in the disassembly that start with `if (!gbNoSound)` so I set it to zero trying to get my sound to work.  I'll put the original code back the way it was in `init.cpp`.

As for my sound issue itself, I've only been able to determine that it's something to do with the `WAVE_init_driver` function.  In the debugger I saw that the `hdidriver` variable was null and that no sound sample objects were allocated.  My code reproduces what the initialization code did, and somehow that allows it to work.  I'm guessing this is Windows 10 specific, or possibly even specific to my machine.  My last change to exit early should allow this to continue working for people whose sound is already working.

src/cpp/shared/sound/cd_music.cpp
```diff
@@ -222,4 +224,46 @@ void soundManager::CDPlay(int track_number, signed int a3, int a4, int a5) {
...
+  fmt.nAvgBytesPerSec = fmt.nSamplesPerSec;
+  fmt.nBlockAlign = 1;
+  fmt.wBitsPerSample = 8;
+  fmt.cbSize = 0;
+  HDIGDRIVER driver = 0;
+  const int status = AIL_waveOutOpen(&driver, 0, 0, &fmt);
+  if (status == 0) {
+    hdidriver = driver;
+    AllocateSampleHandles();
+  }
```
jkoppel: Change this to "} else {"

jkoppel:
The changes you made limit the downside risk, so I can't see a reason not to include this. Make the small formatting change, and then you can merge it yourself.

My guess is that the difference is related to the linker. The way to debug this would be to step through both at the assembly level and look for the difference, which I'll be happy to teach you how to do if you're so interested.

PR #108, opened 2018-03-21
Suppression of Paralyzed Griffin retaliation

cometfractals:
This commits are either good or darn near to it.


jkoppel:
You said that this currently is not working, right? I'm going to close this.

PR #109, opened 2018-04-18
Replacing hard-coded artifact strings with new xml file.

mkristofik:


src/cpp/shared/artifacts.cpp
```diff
@@ -15,437 +19,72 @@
...
-};
+  int TotalArtifacts(const artifacts::artifact_sequence &artifactList) {
+    int numArtifacts = 0;
+    for (const auto &art : artifactList) {
+      const int i = art.id();
+      if (i > numArtifacts && i < NUM_SUPPORTED_ARTIFACTS) {
+        numArtifacts = i;
+      }
+    }
+    return numArtifacts + 1;
```
jkoppel: This will return 1 even if there are 0 artifacts. It also has problems if we make artifacts that skip IDs; perhaps this is misnamed.

src/cpp/shared/artifacts.cpp
```diff
@@ -15,437 +19,72 @@
...
+      desc += "}\n\n";
+      desc += JoinSequence(art.description());
+      descriptions[i] = desc;
+    }
+    gArtifactDesc[i] = &(descriptions[i][0]);
+
+    events[i] = JoinSequence(art.event());
+    gArtifactEvents[i] = &(events[i][0]);
+
+    const auto lvl = static_cast<unsigned char>(std::pow(2, static_cast<int>(art.level())));
```
jkoppel: I don't see how this results in the correct value being assigned.

mkristofik: The level in the xml file is an enum type.  Ultimate artifacts have level 0, major level 1, minor level 2, and treasure level 3.  The game expects numbers of 1, 2, 4, and 8 respectively.  I'll add a comment explaining this better.

src/cpp/shared/artifacts.cpp
mkristofik: Naming things is hard.  I'll make it clearer.

src/cpp/shared/artifacts.cpp
jkoppel: Okay, so it's assigned in order of appearance in the XSD file. Is there any way to make this more clear in the  So, you're saying that 32 is the lowest level of artifact rather than the highest one?

I'm more comfortable with an explicit mapping from level name to integer, decoupling us from the order in the XSD file and making this more explicit.

PR #111, opened 2018-05-31
Add cursed item support to artifacts.xml

mkristofik:
And a couple extras:

- include artifacts.xml in the installer
- removed a weird variable that pointed to the artifact name list, off by one index

### DIFF
--- a/src/cpp/shared/artifacts.cpp
+++ b/src/cpp/shared/artifacts.cpp
@@ -70,6 +70,7 @@ unsigned char gArtifactLevel[NUM_SUPPORTED_ARTIFACTS] = { 0 };
 std::vector<std::string> names;
 std::vector<std::string> descriptions;
 std::vector<std::string> events;
+std::vector<char> isCursed;
 
 
 void LoadArtifacts() {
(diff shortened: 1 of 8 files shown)


src/cpp/shared/artifacts.cpp
```diff
@@ -70,6 +70,7 @@ unsigned char gArtifactLevel[NUM_SUPPORTED_ARTIFACTS] = { 0 };
 std::vector<std::string> names;
 std::vector<std::string> descriptions;
 std::vector<std::string> events;
+std::vector<char> isCursed;
```
jkoppel: Why is this a char?

jkoppel: I mean....why is it not vector<boolean> ?

mkristofik: Sorry, it looks like github munged what I wrote.  I meant that `vector<bool>` is commonly avoided and `vector<char>` is a useful alternative.  I'm using `char` as a 1-byte integer in this case. 
 `vector<int>` would work too but takes up more space.

jkoppel: http://www.codingstandard.com/rule/17-1-1-do-not-use-stdvector/

I see. Add a comment about that, and then merge this in yourself.

mkristofik: My next change already has changes to this file so I'll add it there.

PR #112, opened 2018-06-04
Support new artifacts in the hero edit window.

mkristofik:


src/cpp/editor/events.cpp
```diff
@@ -48,7 +49,8 @@ int buildingIdToIdx[32];
 static const std::wstring GetWC(const char *c) {
 	const size_t cSize = strlen(c) + 1;
 	std::wstring wc(cSize, L'#');
-	mbstowcs(&wc[0], c, cSize);
+	size_t result = 0;
+	mbstowcs_s(&result, &wc[0], cSize, c, cSize);
```
jkoppel: Thanks for this.

src/cpp/editor/events.cpp
```diff
@@ -186,7 +188,7 @@ void InitializeTownEdit(HWND hwnd) {
 	for(int i = 0; i < ELEMENTS_IN(monTypeFields); i++) {
 		SendDlgItemMessage(hwnd, monTypeFields[i], CB_ADDSTRING, 0, (LPARAM)L"-empty-");
 		for(int j = 0; j < GetNumCreatures(); j++) {
-			sprintf(gText, "%s", GetCreatureName(j));
+			snprintf(gText, 64, "%s", GetCreatureName(j));
```
jkoppel: Not a bad habit, although I do view this route as pretty hopeless, given the sheer quantity of sprintf in the original code. Where did the number 64 come from?

src/cpp/editor/hero_edit.cpp
```diff
@@ -1,17 +1,39 @@
...
 
 extern HeroExtra gEditedHeroExtra;
 
-extern int __fastcall EditHeroHandler_orig(tag_message&);
-extern void __stdcall FillInHeroEdit(HeroExtra*);
-
 #define NUM_HERO_PORTRAITS 72
 
+namespace {
+  bool fillInArtifacts = true;
```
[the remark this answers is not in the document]
jkoppel: I see. So, the original code was actually in EditHero, but you moved it to FillInHeroEdit because it was easier than moving EditHero into the Ironfist codebase? Makes sense.

src/cpp/editor/events.cpp
jkoppel: Hmmm. This is essentially assuming that creature names will never grow beyond 64 characters. Probably a safe assumption, but this will be a bug if they do, and one that's impossible to catch.

Why not just use NELEM(gText)?

src/cpp/editor/hero_edit.cpp
```diff
@@ -1,17 +1,39 @@
...
+
+  if (fillInArtifacts) {
+    tag_message artifactEvent;
+    artifactEvent.eventCode = INPUT_GUI_MESSAGE_CODE;
+    artifactEvent.xCoordOrKeycode = GUI_MESSAGE_DROPLIST_ADD;
+    for (int i = MAX_EXPANSION_ARTIFACT + 1; i < NUM_SUPPORTED_ARTIFACTS; ++i) {
+      if (gArtifactNames[i]) {
+        for (int j = 0; j < 3; ++j) {
+          artifactEvent.payload = gArtifactNames[i];
+          artifactEvent.yCoordOrFieldID = j + ARTIFACT_SLOT_1;
```
jkoppel: This is the only thing I'm insisting on. Create a GUIAddDroplistRow function and use it instead, in accordance with http://projectironfist.pbworks.com/w/page/115903900/Working%20with%20the%20GUI

src/cpp/editor/events.cpp
mkristofik: Where is NELEM defined?  I couldn't find it.  I tried `sizeof` but apparently that's not legal on an incomplete type.  If I'm reading the asm correctly, I think the actual size is 300.  I'll use that if you can confirm it for me.

jkoppel: https://github.com/jkoppel/project-ironfist/blob/master/src/cpp/shared/base.h#L12

Looks like it's actually called ELEMENTS_IN

jkoppel: Hmmm....so, we declare gText as an incomplete type. ELEMENTS_IN uses sizeof, so that won't work either. If you really really want to use snprintf, then this means we'll probably have to move the definition of gText into Ironfist (and make sure there's no code that references intermediate addresses). Is it really worth it?

jkoppel:
Good  work, as always.

PR #113, opened 2018-06-06
Add custom creatures to the hero edit window.

mkristofik:


src/cpp/editor/hero_edit.cpp
```diff
@@ -10,10 +11,24 @@ extern HeroExtra gEditedHeroExtra;
 
 namespace {
   bool shouldFillInArtifacts = true;
+  bool shouldFillInCreatures = true;
+
+  std::string Capitalize(const char *name) {
```
jkoppel: I find it a bit weird to accept  char* but return a string. This is a file-private function so it doesn't really matter.

mkristofik: `std::string` has undefined behavior if you pass a `nullptr` to the constructor.  I didn't trust what was in `gArmyNames` so I did it this way to be careful.

mkristofik: And you told me in the past you preferred `std::string` anyway.

jkoppel: Yeah; that's written in the Ironfist guidelines. I'd still generally expect a Capitalize function to take and return the same type.

PR #115, opened 2018-06-22
Add custom artifacts to the map win condition.

mkristofik:


src/cpp/editor/editor.cpp
```diff
@@ -14,12 +15,8 @@
...
 
 void __fastcall ShowErrorMessage(const char *str) {
-	char *buf = (char*)ALLOC(strlen(str) + 1);
-	strcpy(buf, str);
-
-	ShowErrorMessage_orig(buf);
-
-	FREE(buf);
+	std::string buf(str);
+	ShowErrorMessage_orig(&buf[0]);
```
jkoppel: Thanks!

PR #116, opened 2018-06-25
Disabling default map events

jkoppel:


jkoppel:
Example script:

```
function OnLocationVisit(loc, x, y)
  if loc == LOCATION_RESOURCE or loc == LOCATION_CAMPFIRE then
    MessageBox("These resources are cursed. We shouldn't pick them up until we defeat the Evil Necromancer.")
	return true
  elseif loc == LOCATION_TOWN then
    MessageBox("Scoundrel! You are not welcome in this town.")
	return true
  else
    return false
  end
end
```

src/cpp/shared/scripting/callback.h
```diff
@@ -43,4 +51,26 @@ void ScriptCallback(const char * funcName, Args... args) {
...
+
+  if (!LuaGlobalExists(map_lua, funcName)) {
+    return nonstd::optional<Res>();
+  }
+
+  lua_getglobal(map_lua, funcName);
+  ironfist_lua_pushmulti(args...);
+  int size = sizeof...(Args);
+  if (lua_pcall(map_lua, size, 1, 0) != LUA_OK) {
+    lua_pop(map_lua, 1); // Unsure if should have this
```
jkoppel: Waiting on an answer to https://stackoverflow.com/questions/51031385/if-lua-pcall-returns-error-are-there-still-return-values-left-on-the-stack

From a quick glance at the implementation of Lua pcall, I believe this is not needed. However, I don't know enough about the Lua VM to know what exact behavior to look for if this is true vs. false.

src/cpp/shared/scripting/callback.h
```diff
@@ -17,6 +19,12 @@ void ironfist_lua_push(void *arg);
 void ironfist_lua_push(std::string arg);
 void ironfist_lua_pushmulti();
 
+template <typename Res>
+nonstd::optional<Res> GetLuaResult(lua_State *L, int arg);
+
+template <>
+nonstd::optional<bool> GetLuaResult(lua_State *L, int arg);
```
mkristofik: The definition of the bool version of this function should go here too, in the .h file.

src/cpp/shared/scripting/callback.h
```diff
@@ -43,4 +51,26 @@ void ScriptCallback(const char * funcName, Args... args) {
   }
 }
 
+template<typename Res, typename... Args>
+nonstd::optional<Res> ScriptCallbackResult(const char * funcName, Args... args) {
```
mkristofik: Do you want to take `args` as a forwarding reference instead of by value?  Then you can `std::forward` them to `ironfist_lua_pushmulti` and not incur copies.  (And that function should probably take its args by forwarding reference too.)

src/cpp/shared/scripting/callback.cpp
```diff
@@ -18,4 +22,14 @@ void ironfist_lua_push(std::string arg) {
 }
 
 void ironfist_lua_pushmulti() {
+}
+
+template <>
+nonstd::optional<bool> GetLuaResult(lua_State *L, int arg) {
```
mkristofik: `arg` isn't used here.  Do you need it?

src/cpp/shared/lib/optional.hpp
```diff
@@ -0,0 +1,1222 @@
+//
+// Copyright (c) 2014-2018 Martin Moene
+//
+// https://github.com/martinmoene/optional-lite
```
mkristofik: SMH that something 1200 lines long is the "lite" version.  Thanks C++...

Also, just curious, which version of the standard are we targeting here?  `std::optional` is available in C++17, is that possible to use here?

src/cpp/shared/scripting/callback.h
```diff
@@ -43,4 +51,26 @@ void ScriptCallback(const char * funcName, Args... args) {
...
+    return nonstd::optional<Res>();
+  }
+
+  if (!LuaGlobalExists(map_lua, funcName)) {
+    return nonstd::optional<Res>();
+  }
+
+  lua_getglobal(map_lua, funcName);
+  ironfist_lua_pushmulti(args...);
+  int size = sizeof...(Args);
```
mkristofik: const

src/cpp/shared/lib/optional.hpp
jkoppel: We're running on Visual Studio 2015, which has only some of the C++11 features. I don't remember all the barriers to upgrading, but we decided a while ago that it was a no-go.

src/cpp/shared/scripting/callback.h
jkoppel: Thanks for the suggestion. I don't think we'll benefit from the time savings, but I do think using std::forward will make this code less readable to most of the team.

src/cpp/shared/scripting/callback.cpp
jkoppel: Woops; I left in some hardcoded values as I was getting this to work. Fixed.

XT3RM1N8R:
Major issue with the event disable: for events based on objects such as artifacts & resources, armies,  campfires, etc we have no way of making them disappear, afterwards. Additionally, the OnLocationVisit callback occurs before the Hero reaches the tile that such objects are on, meaning that the objects are now an impassible wall.

PR #117, opened 2018-06-26
Allow new artifacts to be generated randomly.

mkristofik:
Tested by placing 36 ART-3 objects on a map, counting repeats, and
making sure the game didn't crash.

Also, silenced a few more compiler warnings.

### DIFF
--- a/src/cpp/shared/game/game.cpp
+++ b/src/cpp/shared/game/game.cpp
@@ -103,7 +103,8 @@ void game::InitNewGame(struct SMapHeader *a) {
 			lastPlayed = read_pref<std::string>("Last Map");
 
 		if (lastPlayed.length() < 20) { // otherwise means no registry keys exist yet
-			strcpy(gMapName, lastPlayed.c_str());  // can't determine the size of gMapName
+			const int mapNameSize = 13;  // DOS 8+3 format
+			strcpy_s(gMapName, mapNameSize, lastPlayed.c_str());
 			strcpy_s(this->mapFilename, sizeof(mapFilename), lastPlayed.c_str());
 		}
 	}
@@ -274,14 +275,14 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int 
 }
 
 void game::InitRandomArtifacts() {
-  memset(artifactGeneratedRandomly, 0, sizeof(artifactGeneratedRandomly));
+  ResetGeneratedArtifacts();
   for (int i = 0; i < MAP_WIDTH; ++i) {
     for (int j = 0; j < MAP_HEIGHT; ++j) {
       const auto &tile = map.tiles[j * MAP_WIDTH + i];
       if (tile.objType == (TILE_HAS_EVENT | LOCATION_ARTIFACT)) {
-        // Don't know why we need to divide by 2, but that's what the decompiled code says.
+        // Each artifact has two tiles. Convert from tile id to the artifact id.
         const int artifactId = static_cast<unsigned char>(tile.objectIndex) / 2;
-        artifactGeneratedRandomly[artifactId] = 1;
+        GenerateArtifact(artifactId);
       }
     }
   }
@@ -300,10 +301,10 @@ int game::GetRandomArtifactId(int allowedLevels, int allowNegatives) {
   const int numArtifacts = (xIsExpansionMap ? NUM_SUPPORTED_ARTIFACTS : MAX_BASE_ARTIFACT + 1);
 
   for (int i = 0; i < numArtifacts; ++i) {
-    if ((gArtifactLevel[i] & allowedLevels) == 0) {
+    if ((gArtifactLevel[i] & allowedLevels) == 0) { // TODO: also test for unused
       continue;
     }
-    if (artifactGeneratedRandomly[i]) {
+    if (IsArtifactGenerated(i)) {
       continue;
     }
     if (i == ARTIFACT_SPELL_SCROLL) {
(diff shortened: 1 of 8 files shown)


src/cpp/shared/game/game.cpp
```diff
@@ -102,8 +103,8 @@ void game::InitNewGame(struct SMapHeader *a) {
 			lastPlayed = read_pref<std::string>("Last Map");
 
 		if (lastPlayed.length() < 20) { // otherwise means no registry keys exist yet
-			strcpy(gMapName, lastPlayed.c_str());
-			strcpy(this->mapFilename, lastPlayed.c_str());
+			strcpy(gMapName, lastPlayed.c_str());  // can't determine the size of gMapName
```
jkoppel: All filenames in HoMM are DOS 8+3 format, so all filename buffers are 13 bytes.

src/cpp/shared/game/game.h
```diff
@@ -139,7 +140,7 @@ class game {
 	char relatedToHeroForHireStatus[54];
 	mine mines[144];
 	char field_60A6[144];
-	char artifactGeneratedRandomly[103];
+	char artifactGeneratedRandomly[NUM_SUPPORTED_ARTIFACTS];
```
jkoppel: You can't change the size of built-in structures. Exception: if you decompile everything that uses that structure (and structures that use it directly), as we did for resourceManager.

What you can do is what we did for hero spells: replace this 103-byte array with a 4-byte pointer to a dynamically-allocated array + 99 bytes of padding, decompile all functions that access this memory, accept that it's hard to not leak memory.

Yes, this sucks.

The good news: I'm only finding 4 functions that access this array: InitRandomArtifacts, GetRandomArtifactId, LoadGame, and SaveGame. And they're all already decompiled.

src/cpp/shared/game/game.cpp
```diff
@@ -272,6 +273,75 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
 }
 
+void game::InitRandomArtifacts() {
+  memset(artifactGeneratedRandomly, 0, sizeof(artifactGeneratedRandomly));
+  for (int i = 0; i < MAP_WIDTH; ++i) {
+    for (int j = 0; j < MAP_HEIGHT; ++j) {
+      const auto &tile = map.tiles[j * MAP_WIDTH + i];
+      if (tile.objType == (TILE_HAS_EVENT | LOCATION_ARTIFACT)) {
+        // Don't know why we need to divide by 2, but that's what the decompiled code says.
+        const int artifactId = static_cast<unsigned char>(tile.objectIndex) / 2;
```
jkoppel: It's because each artifact has two tiles. tile.objectIndex / 2 converts the tile ID to the represented artifact ID.

src/cpp/shared/game/game.h
mkristofik: D'oh.  I was hoping I could get away with that because the xml format is somewhat agnostic to the size of the array there.  I'll have to work on this to come up with another approach - I'll look at hero spells for an example.

src/cpp/shared/game/game.cpp
```diff
@@ -427,7 +498,7 @@ void game::ProcessOnMapHeroes() {
           randomHero->ownerIdx = mapExtraHero->owner;
           this->relatedToHeroForHireStatus[mapExtraHero->heroID] = randomHero->ownerIdx;
           this->players[randomHero->ownerIdx].heroesOwned[this->players[randomHero->ownerIdx].numHeroes++] = randomHero->idx;
-          if (y > 0 && (this->map.tiles[x + ((y - 1) * this->map.width)].objType) == TILE_HAS_EVENT | LOCATION_TOWN) {
+          if (y > 0 && this->map.tiles[x + ((y - 1) * this->map.width)].objType == (TILE_HAS_EVENT | LOCATION_TOWN)) {
```
jkoppel: Yikes. This was bad. Good catch.

src/cpp/shared/game/game.cpp
mkristofik: I've run across other code that refers to the low tile id and the high tile id.  Do you know/can you explain how that works?

src/cpp/shared/game/game.cpp
```diff
@@ -272,6 +273,75 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
...
+    winConditionArtifact = mapHeader.winConditionArgumentOrLocX - 1;
+  }
+
+  // Cursed artifacts are allowed 70% of the time, every time.
+  const bool allowCursed = (allowNegatives && Random(0, 100) >= 30);
+
+  std::vector<int> choices;
+  const int numArtifacts = (xIsExpansionMap ? NUM_SUPPORTED_ARTIFACTS : MAX_BASE_ARTIFACT + 1);
+
+  for (int i = 0; i < numArtifacts; ++i) {
```
jkoppel: What about ARTIFACT_NONE?

mkristofik: The four NONEs have level "unused" (32), which won't match any of the requested levels in the game.  The callers of the this function only request some combination of treasure, minor, and/or major.

src/cpp/shared/game/game.cpp
jkoppel: Not sure what you're referring to. Example?

src/cpp/shared/game/game.h
jkoppel: Nope. It's because everything in heroes2.asm is using a hard-coded structure offset. When you mess this up (as we have in the past), you get bugs where saying "foo.x = 1" doesn't actually change foo.x. They're quite nasty to debug.

src/cpp/shared/game/game.cpp
mkristofik: Line 28515 of HEROES2W.c, random artifacts get converted into two objects via game::ConvertObject.

src/cpp/shared/game/game.cpp
jkoppel: This is a Level 3 bug ( http://www.pathsensitive.com/2018/01/the-three-levels-of-software-why-code.html ). Compare: https://chromium-review.googlesource.com/c/catapult/+/985025/5/telemetry/telemetry/internal/results/story_run.py#15

We need this to be a function which gets a random valid artifact ID, not a function which gets a random valid artifact ID so long as you happen to only request certain artifact ranges.

src/cpp/shared/game/game.cpp
jkoppel: What you're seeing: It's choosing the artifact ID, replacing the random artifact's main tile with the main tile of the chosen artifact, and doing similar for the tile directly to the left.

jkoppel:
I find it weird to put the randomly-generated artifacts list in artifacts.cpp, as they're a property of the map, not a property of the artifacts. But, I'll merge this nonetheless.

PR #118, opened 2018-07-01
Set Call boundry to GetHero

cometfractals:
Check  GetHero(player,hero) and call error if needed


jkoppel:
Closing because this code has not been tested, as per private discussion.

PR #119, opened 2018-07-02
Add support for campaign-only artifacts.

mkristofik:
Campaign-only artifacts are never randomly generated. They can only be
placed manually or given in events. I included all of the Anduran pieces,
the Sphere of Negation, and the Iron Fist in the initial set of campaign-only
artifacts.


src/cpp/shared/game/game.cpp
```diff
@@ -301,22 +301,10 @@ int game::GetRandomArtifactId(int allowedLevels, int allowNegatives) {
...
       continue;
     }
-    if (IsArtifactGenerated(i)) {
-      continue;
-    }
-    if (i == ARTIFACT_SPELL_SCROLL) {
-      continue;
-    }
-    if (xIsPlayingExpansionCampaign
-      && (i == ARTIFACT_BREASTPLATE_OF_ANDURAN
```
jkoppel: :D

PR #122, opened 2018-07-22
Changing DelayTilMilli to used unsigned; patching "Wait 25 days" bug

jkoppel:
I didn't find an easy way to test  this (see https://stackoverflow.com/questions/727918/what-happens-when-gettickcount-wraps ), so I'm asking for manual inspection.

### DIFF
--- a/src/cpp/shared/game/event_loop.cpp
+++ b/src/cpp/shared/game/event_loop.cpp
@@ -20,3 +20,24 @@ void __fastcall Process1WindowsMessage() {
 }
 
 
+unsigned long UTickCount() {
+  return KBTickCount();
+}
+
+/*
+ * This is a slightly hacky fix to the bug described in https://www.celestialheavens.com/forum/7/16792 ,
+ * where a signed overflow of GetTickCount() can cause the game to wait for up to 25 days
+ * 
+ * My preferred fix would involve eliminating all calls to DelayTilMilli to use DelayTil instead,
+ * as DelayTilMilli can't be sure that you don't really want to wait for 25 days. However, properly
+ * using unsigned ints here means that, in the event that a caller of DelayTilMilli inadvertently passes
+ * in a very small number, it will never wait for very long.
+ */
+void __fastcall DelayTilMilli(long tick) {
+  unsigned long uTick = tick;
+  while (UTickCount() < uTick)
+  {
+    Process1WindowsMessage();
+    PollSound();
+  }
+}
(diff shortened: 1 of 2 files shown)


src/cpp/shared/game/event_loop.cpp
```diff
@@ -20,3 +20,24 @@ void __fastcall Process1WindowsMessage() {
...
+ * This is a slightly hacky fix to the bug described in https://www.celestialheavens.com/forum/7/16792 ,
+ * where a signed overflow of GetTickCount() can cause the game to wait for up to 25 days
+ * 
+ * My preferred fix would involve eliminating all calls to DelayTilMilli to use DelayTil instead,
+ * as DelayTilMilli can't be sure that you don't really want to wait for 25 days. However, properly
+ * using unsigned ints here means that, in the event that a caller of DelayTilMilli inadvertently passes
+ * in a very small number, it will never wait for very long.
+ */
+void __fastcall DelayTilMilli(long tick) {
+  unsigned long uTick = tick;
```
mkristofik: Can you make this `const`?

src/cpp/shared/game/event_loop.cpp
```diff
@@ -20,3 +20,24 @@ void __fastcall Process1WindowsMessage() {
...
+  while (UTickCount() < uTick)
```
mkristofik: What about `static_cast<unsigned long>(KBTickCount())` instead of a new function?  It's ugly, but it makes the key bit (using a big enough unsigned type) obvious.

src/cpp/shared/game/event_loop.cpp
```diff
@@ -20,3 +20,24 @@ void __fastcall Process1WindowsMessage() {
 }
 
 
+unsigned long UTickCount() {
+  return KBTickCount();
+}
+
+/*
+ * This is a slightly hacky fix to the bug described in https://www.celestialheavens.com/forum/7/16792 ,
+ * where a signed overflow of GetTickCount() can cause the game to wait for up to 25 days
```
mkristofik: Do you want to add an upper bound to prevent the game ever hanging forever again?  Is it ever reasonable for `DelayTilMilli` to wait for more than a few seconds?

jkoppel: There were two problems that combined to form these long delays:

The first is the possibility of wraparound.

The second is that some code accidentally put in a very small number into DelayTillMilli.

The end result is waiting a very long time.

With this change, the only way DelayTillMilli could wait a very long time is if the caller passed in a very large number. In which case, either they really did want a very long wait, or it was a mistake, in which case we shouldn't hide it (though I guess we can detect it and exit with an error message). In any case, this change cannot introduce any new very-long-hangs where there were none before.

PR #123, opened 2018-07-30
Added lua hook GetTooltipText

mkristofik:


src/cpp/shared/adventure/adv.cpp
```diff
@@ -194,3 +196,43 @@ int advManager::MapPutArmy(int x, int y, int monIdx, int monQty) {
...
+  const int locationType = mapCell->objType & 0x7F;
+  auto overrideText = ScriptCallbackResult<std::string>("GetTooltipText", locationType, xLoc, yLoc);
+  if (!overrideText || overrideText->empty()) {
+    // Lua error occurred or tooltip text not overridden.
+    QuickInfo_orig(x, y);
+    return;
+  }
+
+  // Ensure the tooltip box is visible on the screen.
+  int pxOffset = 32 * x - 57;
```
jkoppel: Is there a way to make these numbers more understandable?

jkoppel:
I really don't like overloading "" to mean "Don't override"; it makes scripts a little unclear (you'd read  the script and think you were making the tooltip blank) .This is an instance of in-band communication.

Unfortunately, the other solutions I see have their own problems.

src/cpp/shared/scripting/callback.h
```diff
@@ -21,14 +21,11 @@ void ironfist_lua_pushmulti();
...
 /*
  * The implementations use Lua functions like lua_isnumber and lua_tonumber.
- * We need template specialization to implement these.
+ * We need template specialization to implement these, see callback.cpp.
  */
 template <typename Res>
 nonstd::optional<Res> PopLuaResult(lua_State *L, int arg);
 
-template <>
-nonstd::optional<bool> PopLuaResult(lua_State *L, int arg);
```
jkoppel: I'm really surprised that template specializations don't need to be in the header file

src/cpp/shared/adventure/adv.cpp
```diff
@@ -150,34 +150,41 @@ void advManager::DoEvent(class mapCell *cell, int locX, int locY) {
 }
 
 void advManager::HandleSpellShrine(class mapCell *cell, int locationType, hero *hro, SAMPLE2 res2, int locX, int locY) {
+  std::string shrineText;
```
jkoppel: Thanks.

PR #124, opened 2018-08-27
Started work on generalizing data structures for the Cyborg faction.

mkristofik:


src/cpp/shared/town/town.h
```diff
@@ -27,7 +48,8 @@ enum FACTION
 	FACTION_MULTIPLE = 6,
 	FACTION_RANDOM = 7,
 
-	FACTION_CYBORG = 13,
+	FACTION_CYBORG = 12,
+	NUM_FACTIONS
```
jkoppel: NUM_FACTIONS is now misnamed; it now actually represents 1 + the max faction ID (13), rather than the numebr of factions (7).

src/cpp/shared/town/town.cpp
```diff
@@ -12,19 +12,175 @@
...
+  0x0,
+  0x0,
+  0x0,
+  0x0,
+  0x0,
+  0x01F8BF9F  // TODO: decide whether Cyborg creatures have any upgrades
 };
 
+// Leave these undefined initially to make the compiler put them in the
+// same data segment. This ensures the old game code will be able to index
```
jkoppel: All such code was in SetupTowns, which you've fixed, correct? So you don't need to leave these undefined initially, right?

mkristofik: Actually, I do.  There's something about initializing `gTownObjNames` with pointers that it ends up in a different memory segment from `gDwellingType`.  Leaving them both uninitialized was the only way I found to make them appear consecutively.

jkoppel: But the only reason to make them appear consecutively is the OOB accesses of `gTownObjNames`, right? And we can eliminate all such OOB accesses, right?

And yes, that makes it change segments. Unitialized arrays are in the .BSS section (which does not stand for "Better Save Space," but might as well). Initialized arrays are in the .DATA section. Const initialized arrays are in the .RDATA section; however, the version of MS C++ used to compile the original game put these in the .DATA section.

src/cpp/editor/hero_edit.cpp
```diff
@@ -10,6 +10,9 @@ extern HeroExtra gEditedHeroExtra;
 #define NUM_HERO_PORTRAITS 72
 
 namespace {
+  const int ARTIFACT_DROPDOWN_1 = 308;
```
jkoppel: I really like your use of namespaces.

PR #132, opened 2018-10-28
More placeholder stuff for Cyborg faction

mkristofik:


src/cpp/shared/town/buildings.h
```diff
@@ -1,6 +1,16 @@
 #ifndef BUILDINGS_H
 #define BUILDINGS_H
 
+struct building_cost {
```
jkoppel: Wow! I must have created this structure a long time ago, before I discovered the version of HEROES2W.exe with the debug symbols.

This should probably be renamed something like "SBuildingCost," in keeping with naming conventions.

src/cpp/shared/adventure/hero.cpp
```diff
@@ -49,7 +49,7 @@ signed __int8 captainStats[MAX_FACTIONS][4] = {
   {0, 0, 0, 0},
   {0, 0, 0, 0},
   {0, 0, 0, 0},
-  {1, 1, 1, 1}
+  {1, 1, 1, 1}  // TODO: is Cyborg a might or magic hero?
```
jkoppel: That's a good question.

They should probably be considered a magic hero, since we're developing a lot of custom spells for them.

src/cpp/shared/adventure/adv.h
```diff
@@ -180,7 +180,8 @@ enum HERO_FLAGS {
   HERO_AT_SEA = 0x80
 };
 
-char cHeroTypeInitial[];
+extern char cHeroTypeInitial[];
+extern signed __int8 captainStats[][4];
```
jkoppel: Do we not have a constant for the number of primary skills?

src/cpp/shared/town/buildings.cpp
```diff
@@ -99,4 +99,251 @@ char * GetIronfistDwellingName(int faction, int dwelling) {
   }
 
   return "INVALID DWELLING";
-}
\ No newline at end of file
+}
+
+extern BUILDING_CODE gBuildingsToDraw[MAX_FACTIONS][BUILDING_MAX] = {
```
jkoppel: Why is this extern?  I don't know what it means to define a variable but also declare it with extern linkage.

src/cpp/shared/town/buildings.h
mkristofik: Sure I can rename it.  This name matches the type in the asm file - not sure if that mattered.

src/cpp/shared/town/buildings.cpp
mkristofik: Confirmed that `extern` isn't required here and removed it.

src/cpp/shared/town/buildings.h
jkoppel: Cool. Please do.

I spent a lot of time 2010-2012 reverse-engineering the HoMM II source code. When I discovered structures, I gave whatever names I saw fitting to them using my own naming convention.

In 2012, I discovered on GrayFace's website a version of the HoMM II binary with debug symbols. I successfully changed the HoMM II decompilation over to use them in June 2012, with the help of the bindiff tool. Upon doing so, we learned the names that the original developers used for all their classes and structures. Over the rest of the year, we converted all the names in IDA (where the names in the disassembly come from) to match the ones in the binary.

However, a few names slipped through the cracks, and this is one of them. If we were to create this structure today, the assembly name would indeed follow the naming convention of the original game.

PR #134, opened 2018-12-31
Treants should not be marked as archers

jkoppel:


jkoppel:
@Kert 

https://github.com/jkoppel/project-ironfist/commit/7123040821f28241f545e4d69125427eebfc37f1#diff-934268e322fb41e22bc50e1696f0a614R388

Any idea what happened?

Kert:
@jkoppel Probably it's a left-over from testing, sorry. Just removing projectile definition for treants in xml will work I guess

PR #135, opened 2019-01-01
Many scripting bugfixes

jkoppel:
*) Adding consistency to player scripting functions giving objects vs. numbers
*) Now properly passing deepbinding objects to callbacks
*) Refactored the C++-side handling of deppbinding
*) Updated Sorrow's End script for deepbinding
*) Rejuvenated Last Stand code; it works again
*) Some installer tweaks


jkoppel:
@mkristofik Please have a look at the template stuff
@Kert Please have a look at the scripting stuff
@XT3RM1N8R You can also have a look.

src/cpp/shared/scripting/deepbinding.cpp
```diff
@@ -0,0 +1,43 @@
...
+  lua_pushstring(L, "ptr");
+  lua_pushinteger(L, ptrAddr);
+  lua_settable(L, -3);
+  lua_getglobal(L, metatable_name.c_str());
+  lua_setmetatable(L, -2);
+}
+
+
+void deepbound_push(lua_State *L, deepbind<army*> ptrAddr) {
+  make_deepbound_object(L, (int)ptrAddr.get(), "battleStack_mt");
```
mkristofik: What is the new `deepbind` class buying you here?  Couldn't you just take an `army*` argument and `reinterpret_cast<int>` to get the integer that lua needs?

jkoppel: The purpose is to signal to the compiler which version of deepbound_push to invoke, and to get a compiler error if no special version has been written. Using reinterpret_cast fails both goals.

jkoppel:
This diff now opportunistically contains some installer improvements

maps/Behemoth.MX2.lua
```diff
@@ -0,0 +1,4 @@
+SetDaysAfterTownLost(GetPlayer(0), 30);
```
Kert: What's the puprose of this? I am not sure when will this is called since this is not under any callback, but it is supposed to be under one of them

maps/Behemoth.MX2.lua
```diff
@@ -0,0 +1,4 @@
+SetDaysAfterTownLost(GetPlayer(0), 30);
+function OnMapStart()
+GrantArmy(GetHero(GetPlayer(0), 0), CREATURE_CYBER_BEHEMOTH, 1);
```
Kert: Indentation?

PR #136, opened 2019-01-03
ScriptCallbackResult now silently returns a nothing value when there is no result

jkoppel:
Addresses http://heroes2.forumactif.com/t660-ironfist-ver-1-3-0-bug-with-fonction-onlocationvisittype-x-y#7080


PierceLBrooks:
Aside from the install conflict, this commit's code changes are acceptable to merge into the master branch.

PR #138, opened 2019-01-08
Fix crash when shooting projectile at creature in bottom row

jkoppel:


jkoppel:
@XT3RM1N8R @tornado711 Lots of stuff for you to learn here. Notice how I identified and named useful operations, and how this made the code cleaner. Notice the use of blocks to make the earlier portions more skimmable, and the named arguments to prevent passing in integers in the wrong order.

Now, if you know that HoMM runs at 640x480, how easy is the bug to spot?

@Kert So, this was actually an original game bug which was merely hidden by having a low value of halfProjHeight. You're absolved of all responsibility.

src/cpp/shared/combat/army.cpp
```diff
@@ -1343,83 +1345,100 @@ void SpecialAttackGraphics(army *attacker, army *target) {
...
-  int projStartY = attacker->frameInfo.projectileStartOffset[1][1] + gpCombatManager->combatGrid[attacker->occupiedHex].occupyingCreatureBottomY;
-  int totXDiff = targMidX - projStartX;
-  char firingLeft = 0;
-  if (targMidX - projStartX < 0) {
-    firingLeft = 1;
-    totXDiff = -totXDiff;
+  int targMidX, targMidY;
+  int totXDiff, yDiff;
+  char firingLeft;
+  {
```
Kert: Why do you use scope when there is no conditional statement?

src/cpp/shared/combat/army.cpp
```diff
@@ -1343,83 +1345,100 @@ void SpecialAttackGraphics(army *attacker, army *target) {
...
-      spriteIdx = attacker->frameInfo.numMissileDirs - 1;
-    else
-      spriteIdx = i - 1;
-  } else {
-    if (yDiff <= 0)
-      spriteIdx = 0;
-    else
-      spriteIdx = attacker->frameInfo.numMissileDirs - 1;
-    angleDeg = (double)(yDiff <= 0 ? 90 : -90);
+  {
```
Kert: Same here

src/cpp/shared/combat/army.cpp
```diff
@@ -1343,83 +1345,100 @@ void SpecialAttackGraphics(army *attacker, army *target) {
...
+    if (totXDiff) {
+      float slope = (double)-yDiff / (double)totXDiff;
+      angleDeg = atan(slope) * 180.0 / 3.14159;
+      int i;
```
Kert: This can be moved inside the loop I think

src/cpp/shared/combat/army.cpp
```diff
@@ -1343,83 +1345,100 @@ void SpecialAttackGraphics(army *attacker, army *target) {
...
+      for (i = 1;  attacker->frameInfo.numMissileDirs > i &&
+        (*(float *)((char *)&attacker->frameInfo.projectileStartOffset[i + 2] + 1) +
+          attacker->frameInfo.projDirAngle[i]) / 2.0 >= angleDeg;  ++i)
+        ;
+      if (attacker->frameInfo.numMissileDirs <= i)
+        spriteIdx = attacker->frameInfo.numMissileDirs - 1;
+      else
+        spriteIdx = i - 1;
+    }
+    else {
```
Kert: Broken indentation

src/cpp/shared/combat/army.cpp
```diff
@@ -1343,83 +1345,100 @@ void SpecialAttackGraphics(army *attacker, army *target) {
...
+      angleDeg = (double)(yDiff <= 0 ? 90 : -90);
+    }
   }
 
   int attackDirectionAnimationIdx;
-  if (angleDeg <= 25.0) {
-    if (angleDeg <= -25.0) {
-      attacker->animationType = ANIMATION_TYPE_RANGED_ATTACK_DOWNWARDS;
-      attackDirectionAnimationIdx = 2;
+  {
```
Kert: Once again a new scope

src/cpp/shared/combat/army.cpp
```diff
@@ -1343,83 +1345,100 @@ void SpecialAttackGraphics(army *attacker, army *target) {
...
-      * gfCombatSpeedMod[giCombatSpeed]
-      / (double)attacker->frameInfo.animationLengths[attacker->animationType]);
-  }
-  attacker->animationFrame = attacker->frameInfo.animationLengths[attacker->animationType] - 1;
-  int projIconWidth = 100;
-  int projIconHeight = 100;
-  int v35 = 31;
+  }
+
+  DrawShootingAnimation: {
```
Kert: A goto label without goto? :thinking:

src/cpp/shared/combat/army.cpp
```diff
@@ -1343,83 +1345,100 @@ void SpecialAttackGraphics(army *attacker, army *target) {
...
+    for (attacker->animationFrame = 0;
+         attacker->frameInfo.animationLengths[attacker->animationType] > attacker->animationFrame;
+         ++attacker->animationFrame) {
+
```
Kert: Useless new line?

src/cpp/shared/combat/army.cpp
jkoppel: This is a technique I've become a fan of. It's a nice way to separate sections of the function without breaking things out into new functions. 

I don't remember where exactly I learned it, but http://number-none.com/blow/john_carmack_on_inlined_code.html played an influence.

src/cpp/shared/combat/army.cpp
jkoppel: It's because otherwise the next line would crowd into this for loop header, making it less clear where the loop body begins.

src/cpp/shared/combat/army.cpp
jkoppel: Yep, exactly. It's a label for this block of code. I'd label the other bare-braced blocks if I had a good name for them.

src/cpp/shared/combat/army.cpp
jkoppel: Is there a need though?

src/cpp/shared/combat/army.cpp
Kert: Doesn't this _ask_ to be moved into a separate function then? Same for scopes I mentioned above

jkoppel: This one looks like it can be pretty manageably. The ones above less so because they have multiple outputs. Identifying a block like this and labeling it is pretty much always a win. It takes more thought to decide whether splitting something into a function is a good idea because there are more tradeoffs.

PR #145, opened 2019-06-07
Make flag render behind hero

jkoppel:
@Kert  Please review

### DIFF
--- a/src/asm/editor_imports.inc
+++ b/src/asm/editor_imports.inc
@@ -156,3 +156,13 @@ IMPORT_?FillInWinCondition@@YIHH@Z = 1
 
 ;;spells
 IMPORT_?gSpellNames@@3PAPADA = 1
+
+;;towns
+IMPORT_?gWellExtraNames@@3PAPADA = 1
+IMPORT_?gSpecialBuildingNames@@3PAPADA = 1
+IMPORT_?gBuildingInfoSpecial@@3PAPADA = 1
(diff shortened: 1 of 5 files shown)


PierceLBrooks:
I'm unfortunately not seeing the hero at all right now. I repackaged the AGG assets and such, but with no luck. Here is a personal test merge branch:
https://github.com/PierceLBrooks/project-ironfist/commit/c98fe6da7f79f14ccfaefaf93b522ca84847c7b6

PierceLBrooks:
Should ironfist_save::hero_t::relatedToUnknown() be refactored to ironfist_save::hero_t::directionFacing() ?

jkoppel:
Eventually, but it should wait for other schema changes.

Currently, every time we change the save-game schema, it will *break* all existing savegames.

jkoppel:
So, have you reviewed the code and found no other concerns?

Kert:
Are you sure you want this to be done before adding cyborg specific adventure map hero graphics?
Because it requires adding cbrg32.icn (like wzrd32.icn e.t.c.) and they are stored in advManager::heroIcons[6] array.
What I mean is that if we want to add separate graphics we'll need to import the whole advManager::DrawCell and advManager::DrawCursor functions + advManager::Open (this one is easy) and store this data somewhere else.
We can't use the current code in this branch for that since it draws stuff on top of what was drawn by original functions.
I suggest that we implement the feature of rendering flag behind hero along with cyborg heroes and maybe even exactly for them.

jkoppel:
Okay. Very sad.

PR #146, opened 2019-06-14
Fixed MapPutArmy

jkoppel:


jkoppel:
@Kert  I suspect that you only tested this function on symmetric coordinates like (10,10).

PR #148, opened 2019-06-21
"SetInclinedToJoin" & "GetInclinedToJoin" script call implementations

PierceLBrooks:
This also includes a "OnMonsterInteract" callback.

### DIFF
--- a/src/cpp/shared/adventure/adv.cpp
+++ b/src/cpp/shared/adventure/adv.cpp
@@ -314,3 +314,67 @@ void advManager::QuickInfo(int x, int y) {
   QuickViewWait();
   gpWindowManager->RemoveWindow(&tooltip);
 }
+
+void advManager::PlayerMonsterInteract(mapCell *cell, mapCell *other, hero *player, int *window, int a1, int a2, int a3, int a4, int a5) {
+	int creat = cell->objectIndex;
+	int amt = cell->extraInfo;
+	int x;
+	int y;
+	if (!(cell->objType & LOCATION_ARMY_CAMP)) {
+		this->PlayerMonsterInteract_orig(cell, other, player, window, a1, a2, a3, a4, a5);
+		return;
+	}
+	if (GetMapCellXY(cell, &x, &y)) {
+		ScriptCallback("OnMonsterInteract", x, y);
+	} else {
+		this->PlayerMonsterInteract_orig(cell, other, player, window, a1, a2, a3, a4, a5);
+		return;
+	}
+	if (gpGame->monstersWillJoin[x][y]) {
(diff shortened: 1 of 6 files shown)


src/cpp/shared/adventure/adv.cpp
```diff
@@ -313,4 +313,55 @@ void advManager::QuickInfo(int x, int y) {
...
+	int x;
+	int y;
+	if (cell->objType != (LOCATION_ARMY_CAMP | TILE_HAS_EVENT)) {
+		this->PlayerMonsterInteract_orig(cell, other, player, window, a1, a2, a3, a4, a5);
+		return;
+	}
+	if (GetMapCellXY(cell, &x, &y)) {
+		ScriptCallback("OnMonsterInteract", x, y);
+	}
+	if ((x < 0) && (y < 0) && (x >= gpGame->map.width) || (y >= gpGame->map.height)) {
```
jkoppel: Why are the bounds checks necessary?

src/cpp/shared/adventure/adv.cpp
```diff
@@ -313,4 +313,55 @@ void advManager::QuickInfo(int x, int y) {
...
+		this->PlayerMonsterInteract_orig(cell, other, player, window, a1, a2, a3, a4, a5);
+		return;
+	}
+	int cellIdx = y * gpGame->map.height + x;
+	this->PlayerMonsterInteract_orig(&gpGame->map.tiles[cellIdx], other, player, window, a1, a2, a3, a4, a5);
```
jkoppel: Isn't the first argument just "cell"?

src/cpp/shared/game/game.h
```diff
@@ -162,7 +162,7 @@ class game {
 	char field_660D;
 	char field_660E;
 
-  // New state
+  // New states
```
jkoppel: Please revert this

src/cpp/shared/game/game.cpp
```diff
@@ -214,11 +214,11 @@ void game::PerMonth() {
 }
 
 void game::ResetIronfistGameState() {
```
jkoppel: Revert this as well. I recommend inspecting your git status before making any commits so you don't accidentally commit whitespace changes like this.

src/cpp/shared/game/game.h
PierceLBrooks: Ah, this was left behind from when I was using a boolean checker grid for the WillJoin implementation. That array was declared here in a similar fashion, so it made sense to me at the time to pluralize the comment. I am undoing this now.

src/cpp/shared/adventure/adv.cpp
PierceLBrooks: Yes it is. I am simplifying this now. The overly verbose form was just for sanity's sake. Thanks for pointing it out.

PR #150, opened 2019-07-02
implemented "PlaySound" & "MapFizzle" script functions

PierceLBrooks:


src/cpp/shared/scripting/funcs.cpp
```diff
@@ -15,8 +15,59 @@ extern "C" {
...
+			WaitEndSample(res, res.sample);
+		}
+		free(src);
+		return true;
+	}
+	return false;
+}
+
+static bool CheckLocationItem(mapCell *loc) {
+	if (loc->objType == LOCATION_ANCIENT_LAMP) {
```
jkoppel: Have you tested this? I don't think this works. The high bit is TILE_HAS_EVENT.

Also, a shorter way to test this:

switch (x) {
  case LOCATION_ANCIENT_LAMP:
  case LOCATION_ARTIFACT:
  case LOCATION_RESOURCE
  ...
  case LOCATION_SEA_CHEST:
    return true;
  default:
    return false
}

PierceLBrooks: I think that has been addressed. Weird that it was working for me before, but the expression should now correctly use the "objType" field from the map cell.

PR #151, opened 2019-07-02
Ensured no duplicate map ending callbacks with simple boolean flag

PierceLBrooks:


PierceLBrooks:
The original implementation of game::NextPlayer literally just calls game::CheckEndGame twice, so it seems there is unfortunately no way to overcome this bug without using a boolean flag as I have done already with the present pull request solution.

jkoppel:
I authorize you to merge this PR.

PR #155, opened 2019-07-18
Buffer overflow in getting map name from registry

jkoppel:
@Kert  Do you know what's up with this? I found some uncommitted code in my git stashes.

### DIFF
--- a/src/cpp/shared/game/game.cpp
+++ b/src/cpp/shared/game/game.cpp
@@ -119,8 +119,8 @@ void game::InitNewGame(struct SMapHeader *a) {
 		else
 			lastPlayed = read_pref<std::string>("Last Map");
 
-		if (lastPlayed.length() < 20) { // otherwise means no registry keys exist yet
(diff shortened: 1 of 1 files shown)


Kert:
@jkoppel This is exactly where my game crashes upon loading "LastStand.MX2"
But from the looks of this commit this name will still not be supported (albeit not crashing anymore) since it's too long

jkoppel:
Yeah, I'm really not sure how you manage to load  such a map in the first place.

Any reason not to merge this now?

Kert:
Maybe add support for longer map names? Because it's the only reason I can think of
You can load the map with long names. But when you restart the game the saved name is too long and it crashes during map selection list loading.

jkoppel:
Okay: https://trello.com/c/FkTGs1vf/268-make-the-remember-what-map-was-last-played-feature-work-with-map-filenames-longer-than-13-characters

Merging now.

PR #159, opened 2019-07-22
Implemented Forced Computer Player Chasing Mechanic

PierceLBrooks:


src/cpp/shared/game/save.cpp
```diff
@@ -900,9 +914,40 @@ void game::LoadGame(char* filnam, int newGame, int a3) {
       std::map<std::string, mapVariable> mapVariables;
       for (ironfist_save::save_t::mapVariable_const_iterator it = mp->mapVariable().begin();
         it != mp->mapVariable().end(); it++) {
-        mapVariable *mapVar = new mapVariable;
         std::string mapVariableId = it->id().get();
         MapVarType mapVariableType = StringToMapVarType(it->type());
+		if ((mapVariableType == MAPVAR_TYPE_BOOLEAN) && (!mapVariableId.empty())) {
```
jkoppel: What's with the indentation discontinuity?

src/cpp/shared/game/save.cpp
```diff
@@ -900,9 +914,40 @@ void game::LoadGame(char* filnam, int newGame, int a3) {
...
+		  if ((mapVariableId[0] == '_') && ((it->value().get() == "true") || (it->value().get() == "True") || (it->value().get() == "TRUE"))) {
+			if (mapVariableId.length() > 10) {
```
jkoppel: Is this redundant with the next check?

src/cpp/shared/game/save.cpp
```diff
@@ -900,9 +914,40 @@ void game::LoadGame(char* filnam, int newGame, int a3) {
...
+			  if (mapVariableId.substr(1, 8) == "AICHASE_") {
```
jkoppel: I think this entire block can be replaced with a single sscanf call, or whatever its C++ equivalent is.

jkoppel: If you can't simplify the code, then at least put it in a method. This is a great case of "complicated code that has simple behavior," which makes it a strong candidate for extracting a method.

src/cpp/shared/game/game.cpp
```diff
@@ -291,6 +299,15 @@ void philAI::RedistributeTroops(armyGroup *army1, armyGroup *army2, int a1, int
 	}
 }
 
+int philAI::EvaluateHeroEvent(int a1, int a2, int a3, int a4, int *a5) {
+	hero *src = GetCurrentHero();
+	hero *dst = &gpGame->heroes[a1];
+	if (gpGame->forcedComputerPlayerChases[src->idx][dst->idx]) {
+		return AI_VALUE_CAP;
```
jkoppel: Gorgeous.

src/cpp/shared/game/save.cpp
PierceLBrooks: My logic is to make sure that the format is proper like this: "_AICHASE_X_Y_"

src/cpp/shared/game/save.cpp
PierceLBrooks: This has now been addressed.

src/cpp/shared/game/save.cpp
PierceLBrooks: This has now been addressed.

jkoppel:
This commit arguably made it worse. There is still a big block of complicated logic in the middle of the general map variable code, except now that knowledge of this format has been spread to a third place. This is also the problem that you're doing a lot of work to check if the name can be parsed, and then doing that work again to actually parse it. I've already told you about boolean blindness.

How about: 

int x, y;
char *str = it->value().get().c_str();
if (sscanf(str, "AICHASE_%d_%d", &x, &y)) {
  gpGame->forcedComputerPlayerChases[x][y] = true;
}

40 complicated lines into 5 simple lines.


It sounds like there's not a good alternative for C++ strings:

https://www.quora.com/Is-there-a-C-alternative-to-sscanf

PierceLBrooks:
I have simplified the reading of these values according to your specifications.

jkoppel:
Have you tested this still works?

If so, you may merge.

PierceLBrooks:
Yes this has been tested and still works.
I do not yet have permission as far as I am aware.

PR #168, opened 2019-09-12
Bug fixes

Kert:
Game:
Fixed creatures not moving after casting Slow
Fixed ironfist things like share vision being incorrectly applied
Fixed Captain's Quarters being shown as built already after making another build on the same day
Fixed animation of flying stuttering at the beginning
Remove Shadow Mark from mage guilds
Fixed a problem when special characters (< > & " ') could cause map script loading from savegame to fail
Fixed OnMapVictory being called before the campaign scenario map actually loaded
Fixed potential OOB access of iGetSSByAlignment in hero::CheckLevel
Fixed phantom hero bug (same hero owned by multiple players)
Added debug option to add cyborg creatures to army

Visual Studio Project:
Added missing msg.h to vs project headers
Fixed building Release configuration for both main game and editor


src/cpp/shared/combat/army.cpp
```diff
@@ -2279,11 +2278,10 @@ signed int army::SetSpellInfluence(int effectType, signed int strength) {
       this->frameInfo.stepTime = (signed __int64)((double)this->frameInfo.stepTime * 0.65);
       break;
     case EFFECT_SLOW:
-      this->CancelIndividualSpell(0);
-      v4 = (signed int)this->creature.speed + 1;
-      this->creature.speed = ((signed int)v4 - v4) >> 1;
-      if (this->creature.creature_flags & 2)
-        this->creature.creature_flags -= 2;
+      this->CancelIndividualSpell(EFFECT_HASTE);
+      this->creature.speed = (this->creature.speed + 1) / 2;
```
jkoppel: Wow! I had always thought that Slow was a flat -2 to speed. Just tested in the original game; it's not.

src/cpp/shared/town/town.cpp
```diff
@@ -1210,9 +1210,10 @@ void townManager::SetupCastle(heroWindow *window, int a3) {
 
   if(imgIdx == -1) {
     GUIRemoveFlag(casWin, 1102, 4);
-    GUISetImgIdx(casWin, 1102, imgIdx);
-  } else
-    GUIAddFlag(casWin, 1102, 4);
+  } else {
+	GUISetImgIdx(casWin, 1102, imgIdx);
```
jkoppel: @Kert  Can you explain how this code works?

src/cpp/shared/adventure/adv.cpp
```diff
@@ -786,6 +786,15 @@ bool GetMapCellXY(mapCell* cell, int* x, int* y) {
 	return false;
 }
 
+#include <iostream>
+
+char TileToBitmap_orig(tileset *tiles, int idx, bitmap *targ, int x, int y) {
+  if(x == -1) { // if this is removed the release build crashes
+    std::cout << idx << std::endl;
```
jkoppel: Any idea why this works?

jkoppel:
I left a couple questions.

Before we merge, we either need to redo the temporary fix of TileToBitmap, or I'll need to inspect further to ensure no functional degradation.

src/cpp/shared/adventure/adv.cpp
Kert: I think that was false positive from me since it stopped working later (I think I just used debug build accidentally)

src/cpp/shared/town/town.cpp
Kert: It's a typo. Previously it was setting imgIdx to a widget when it was -1, which is not right obviously

Kert:
> I left a couple questions.
> 
> Before we merge, we either need to redo the temporary fix of TileToBitmap, or I'll need to inspect further to ensure no functional degradation.

I can only say that I've tested the game by dragging map overview everywhere on a couple of big maps after the fix (using _orig func) and didn't have any crashes.

jkoppel:
I unfortunately cannot accept a commit that disables a large amount of code which was presumably written to contain useful features (such as the flag rendering behind the hero)..

I recommend reverting the changes  to TileToBitmap, and declaring that Release mode does not yet work. That will enable us to merge this and all branches depending on it in the meantime.

Kert:
> I unfortunately cannot accept a commit that disables a large amount of code which was presumably written to contain useful features (such as the flag rendering behind the hero)..
> 
> I recommend reverting the changes to TileToBitmap, and declaring that Release mode does not yet work. That will enable us to merge this and all branches depending on it in the meantime.

It disables only one render pass not the whole func

jkoppel:
If there are no intentional changes to DrawCell compared to the original code in this block, then you should just remove everything after the return.

Kert:
What if we this code will be needed some day? Will have to be "reversed" again

jkoppel:
Yes. Commented out code is not for preserving old functionality.  Dead code is even more not for preserving old functionality. That's what Git is for. You can leave a comment there pointing the reader to the git commit if you want to leave something in the code.

PR #170, opened 2019-09-18
Campaigns from files

Kert:
Created a system of loading custom campaigns from files (*.cmp)
These files are xml-s in disguise
Total requirements for a new campaign to work:
- *.cmp file
- 3 frames in X_CMPEXT.ICN - frame ids starting from (3 +campaignID * 3) e.g. for a campaign with id 4 new frames should be 0015,0016,0017
- 1 frame in X_CMPHDR.ICN - frame ids = campaign id
- New ICN file with one frame X_TRACK*.ICN where * = campaign id
- Campaign map files CAMP*_01.HXC, CAMP*_02.HXC e.t.c. where * = campaign id


jkoppel:
Note to self that I have reviewed everything here except campaign.cpp itself.

src/cpp/shared/campaign/campaign.cpp
```diff
@@ -228,6 +175,162 @@ void ExpCampaign::InitNewCampaign(int var) {
...
+  gpGame->numPlayers = gpGame->mapHeader.numPlayers;
+  gpGame->NewMap(gMapName);
+  playerData *player0 = &gpGame->players[0];
+  char heroIdx = player0->heroesOwned[0];
+  hero *hero0 = &gpGame->heroes[heroIdx];
+  
+  CAMPAIGN_CHOICE_TYPES choiceType = (CAMPAIGN_CHOICE_TYPES)campChoice->type;
+  if(choiceType < NUM_CAMPAIGN_CHOICE_TYPES) {
+    switch(choiceType) {
+      case 0u:
```
jkoppel: This needs to be an enum.

src/cpp/shared/campaign/campaign.cpp
```diff
@@ -228,6 +175,162 @@ void ExpCampaign::InitNewCampaign(int var) {
...
+        case 6:
+        case 7:
+          continue;
+      }
+    }
+  }
+  gbRetreatWin = 1;
+}
+
+void ExpCampaign::Autosave() {
```
jkoppel: What was changed in this?

src/cpp/shared/campaign/campaign.cpp
```diff
@@ -400,60 +454,243 @@ void ExpCampaign::ShowInfo(int inGame, int unused) {
...
+  "Sphere of Negation"
+};
+
+std::vector<std::string> xCampaignShortSkillLevel = {
+  "Bas.",
+  "Adv.",
+  "Exp."
+};
+
+void ExpCampaign::UpdateInfo(int drawWindow) {
```
jkoppel: What exactly does this do, and when is it called?

In particular, why didn't it need to be changed for the Cyborg campaign previously?

src/cpp/shared/campaign/campaign.cpp
Kert: string parameter is an item from std::map

src/cpp/shared/campaign/campaign.cpp
Kert: Previously it was using original game array accesses. Now it uses our own (std::map items mostly)

src/cpp/shared/campaign/campaign.cpp
Kert: fixed

PR #171, opened 2019-09-19
Load/Save campaign heroes between scenarios

Kert:
It's possible now to control which hero data transfers to the next scenario by adding a few lines to a corresponding campaign file (*.cmp)
Data transfered:
- Primary/secondary skills
- Spells learned
- Experience

Examples:
`<saveHero scenarioID="0" playerID="0" ownedHeroID="0"/>`
This will save the first owned hero from player 0 to memory on the first campaign scenario

`<loadHero scenarioID="1" playerID="0" ownedHeroID="0"/>`
This will load the first saved hero in memory to the first hero owned by player 0

It's possible to transfer multiple heroes between scenarios but you must follow the same order for saving and loading, e.g.:
`<saveHero scenarioID="0" playerID="0" ownedHeroID="0"/>`
`<saveHero scenarioID="0" playerID="0" ownedHeroID="1"/>`
`<loadHero scenarioID="1" playerID="0" ownedHeroID="0"/>`
`<loadHero scenarioID="1" playerID="0" ownedHeroID="1"/>`


jkoppel:
Nice!

Can I see a video of this?

src/cpp/shared/expansions.h
```diff
@@ -12,7 +12,26 @@ class combatExtra {
         stackExtra stack;
 };
 
+class campaignExtra {
```
jkoppel: This is a brand new class, not present in the original game, right?

jkoppel:
I approve this pending seeing it run, and pending upstream branches.

src/cpp/shared/expansions.h
Kert: Yes. It's a new one

PR #176, opened 2019-09-25
Hotfixes

Kert:
Fixed possible unitialized variable use
Fixed going into expansion campaign crash if AppData registry entry is not pointing to game's folder

### DIFF
--- a/src/cpp/shared/game/game.cpp
+++ b/src/cpp/shared/game/game.cpp
@@ -984,15 +984,6 @@ void game::GiveTroopsToNeutralTown(int castleIdx) {
 }
 
 void game::ProcessOnMapHeroes() {
-  int faction;
-  signed int randomHeroIdx;
-  mapCell *loc;
-  hero *randomHero;
-  char heroExists[MAX_HEROES];
-  int ppMapExtraHeroIdx;
-  HeroExtra *mapExtraHero;
-  char isJail;
-
   struct heroData {
     FACTION faction;
     bool randomizable;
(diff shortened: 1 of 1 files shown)


jkoppel:
On bug 1: Can you explain what was broken and what your change does?
On bug 2: What happens if you run the game from a different folder? (E.g.: In the command line, from a different directory, do something like "..\HoMM2\ironfist.exe".)

Kert:
> On bug 1: Can you explain what was broken and what your change does?
> On bug 2: What happens if you run the game from a different folder? (E.g.: In the command line, from a different directory, do something like "..\HoMM2\ironfist.exe".)

1. randomHeroIdx is not always initialized when the code gets to those lines
2. Error loading creatures.xml

jkoppel:
1. What's the relation between randomHeroIdx and mapExtraHero->heroID ?
2. You're saying...you get that bug both in the old and new versions?

Kert:
> 1. What's the relation between randomHeroIdx and mapExtraHero->heroID ?
> 2. You're saying...you get that bug both in the old and new versions?
1. mapExtraHero->heroID is the actual portrait to be used
randomHeroIdx is heroes[] index which is most of the time the same as portrait ID, except when portrait ID is >= 54 (NUM_HEROES)
2. Your situation described was never in our code. I think even original game would break if run from another folder through command line. Who would do this anyway?!

jkoppel:
Okay; I was just asking that, as your code is assuming the correct current directory.

I actually was able to run HEROES2W.exe from a different folder, but not Ironfist. There is no additional harm from merging this.

PR #179, opened 2019-09-30
Winnable prototypes of map 2, 3 and 4

Kert:


jkoppel:
I will test. Put these in Dropbox as well.

jkoppel:
I tried to test.

Either the map 2 script didn't run, or the AI player died by virtue of running out of days, and then I was unable to win and proceed to the next scenario. Also, the castle should start neutral. Because it started off under the player's possession, the callback did not run.

Kert:
> I tried to test.
> 
> Either the map 2 script didn't run, or the AI player died by virtue of running out of days, and then I was unable to win and proceed to the next scenario. Also, the castle should start neutral. Because it started off under the player's possession, the callback did not run.

You probably haven't updated the map in your build folder, because I specifically changed the castle to not be owned by human player in the updated map.

jkoppel:
Hmmm.

I copy+pasted them from the Ironfist maps folder to the HoMM II MAPS
folder. Could you update the installer because we'll need to do that
eventually, and in case I made a mistake?

Kert:
> Hmmm. I copy+pasted them from the Ironfist maps folder to the HoMM II MAPS folder. Could you update the installer because we'll need to do that eventually, and in case I made a mistake?

Updated.
Just to be sure, are you testing scenario1 which is second campaign scenario which is CAMP5_02?

jkoppel:
CAMP5_02 looks good

CAMP5_03:  Set the human player to blue (the guy in the top left).
CAMP5_04: Looks good

jkoppel:
Also, I'm not seeing the game automatically create a save upon finishing a mission. What's up with that?

Kert:
> Also, I'm not seeing the game automatically create a save upon finishing a mission. What's up with that?

At any mission? It autosaves to CYB_1, CYB_2 e.t.c...

jkoppel:
I see it now. Don't know why I didn't notice them before.

jkoppel:
Reviewed. Still not winnable. I need to check with Abbas on how it's supposed to work.

jkoppel:
Delete the color tents.

Kert:
> Delete the color barriers.

On which map?

jkoppel:
Map 2, which I believe is the only one that has them.

jkoppel:
Tested. 

I now see that blocking  the orange player from getting the boat was critical for ensuring the winnability of this map. We need some way to ensure the player has a boat to reach the final area.
 
Option 1) Reinstate the one color barrier needed to block the Orange player  from nabbing the boat, and then place the tent near the player's starting point.
Option 2) Add a scripting call to make a boat appear, and then create a scripted event to makee a boat appear upon capture of the Barbarian castle
Option 3) Move the Barbarian castle adjacent the water

If you think you can do option 2 in under 1 hour, please do it; otherwise, pick a different one and do it.

jkoppel:
Still waiting  on blocking the orangeg player from getting the boat in CAMP05_3

Has  CAMP05_5 been touched? It's not supposed to be winnable yet, right?

jkoppel:
This branch doesn't have hero saving, right?

Kert:
> This branch doesn't have hero saving, right?
Map 2 loads the saved hero from Map 1

Kert:
> I now see that blocking the orange player from getting the boat was critical for ensuring the winnability of this map. We need some way to ensure the player has a boat to reach the final area.

Which boat? I can see one near barbarian castle (top right) and one near wizard castle (bottom center). Atleast the latter should be 99% sure available for the player
Alternative additional option could be giving a player boat summon spell/give spell scroll. This would probably make more sense than spawning a boat out of the blue

Kert:
> Has CAMP05_5 been touched? It's not supposed to be winnable yet, right?

Should be winnable. First commit in this PR was related to this

jkoppel:
"Map 1 loads the saved hero from Map 0. Not sure this or the other map branch has this"

Yeah, pretty sure it's not in this branch; I don't even see the XML file.

jkoppel:
"Should be winnable. First commit in this PR was related to this"

How? I see a single enemy castle. The game doesn't end when I defeat it.  What's the win condition?

jkoppel:
"Alternative additional option could be giving a player boat summon spell/give spell scroll and placing a boat in the middle of the sea. This would probably make more sense than spawning a boat out of the blue"

Hmmm....how about we make the hero learn Summon Boat after conquering the Warlock castle in the south?

Kert:
> "Should be winnable. First commit in this PR was related to this"
> 
> How? I see a single enemy castle. The game doesn't end when I defeat it. What's the win condition?

Defeat all heroes & capture all castles

jkoppel:
@Kert The change made doesn't guarantee the map winnable, as the player might lose the hero who captured that castle. 

But....the player already loses if John Carver is defeated, right? If you give the spell to him, or to all the player's heroes, that would work.

I'm still worried it may be possible for a hero in a boat to die, and make this map unwinnable if that happens. There still needs to be a source of new boats. Ideas?

jkoppel:
> Defeat all heroes & capture all castles

Quite confident I did. However, after finishing Map 5 and this PR, there will be a lot more requests on map, 4, so it's a moot point.

jkoppel:
I take back what I said about Map 4. It is winnable indeed.

jkoppel:
Ensure Map 2 is winnable, and I'll merge this PR.

I'm about to go design some encounters for Map 4.

Kert:
@jkoppel Which one do you think is the best one:
a)
1. Place multiple boats in a distant unreachable bottom left corner
2. Grant summon boat spell all the time any player hero enters that warlock castle
b)
Place a portal near the top right castle that leads to the lava terrain
c)
Make a new spell that summons a boat even when there are none on the map. I do not suggest altering existing spell because this is confusing for the player

jkoppel:
Doing some thinking:


c) is complicated, b) removes the aquatic nature, and I expect a lot of players to not realize they can use summon boat and get frustrated for (a).

I think the ideal solution is to have a shipyard building on the coast, as a replenishable supply of boats.


Alas, that exists in HoMM III but not II.

The simplest thing to do would be to make the top-right town coastal.

More complicated: Make the shipyard only available to the human player, or make it initially non-coastal but have a "canal" be dug to the castle when it's captured by the human player.

I'll take any of these.

Kert:
I'll make it coastal and in future if you want we can make a shipyard structure as a separate task. Do you agree?

jkoppel:
Hmmm. I am gravitating toward this as the correct solution. Yes, I agree.

We haven't added a new type of adv map location before. Are you confident you can do this?

jkoppel:
Anyway,, proceed on makingg it coastal.

Kert:
> We haven't added a new type of adv map location before. Are you confident you can do this?
I expect it would take more time than I think but ofc it should be doable with our current system

Kert:
coast

jkoppel:
Good job, Sergei!

PR #180, opened 2019-10-02
Set town faction

Kert:


jkoppel:
http://puu.sh/EnXaT/8a57739e32.mp4

Nice work.

I want to see what happens when a town with a building gets changed to one without, and then back.

E.g.: Barbarbarian with Tavern to Necromancer and back
E.g.: Warlock with Black Tower to Sorceress and back.
E.g.: Sorceress with everything to Warlock.

Also, today I learned that the Necromancer Shrine only exists in expansion maps. WTF? I want to see the code that does this.

jkoppel:
(I see the code is very explicit about this, but I still want to see it working.)

src/cpp/shared/town/town.cpp
```diff
@@ -1343,4 +1343,73 @@ int __fastcall CanBuild(town *twn, int building) {
...
+      if(downgradedBldMask & eligibleBuildMask) {
+        this->buildingsBuiltFlags |= downgradedBldMask;
+        this->numCreaturesInDwelling[dwellingIdx] = numCreatures;
+      }
+    }    
+  }
+  
+  // Now we set the offset of graphics for requested faction
+  if(faction == FACTION_CYBORG)
+    faction = (FACTION)6;
```
jkoppel: This code here changes the interpretation of the "faction" variable for a total of one line.

May as well do something like:

if (faction == FACTION_CYBORG) {
  imgOffset += 6;
} else {
  imgOffset += faction;
}

This preserves the meaning of the "faction" variable.

src/cpp/shared/town/town.cpp
```diff
@@ -1343,4 +1343,73 @@ int __fastcall CanBuild(town *twn, int building) {
...
+      continue;
+
+    // Remove building that can't be built
+    this->buildingsBuiltFlags &= ~(bldMask);
+
+    // Downgrade buildings if upgraded ones don't exist in new faction
+    int dwellingIdx = bldIdx - 19;
+    int numCreatures = this->numCreaturesInDwelling[dwellingIdx];
+    if(bldIdx == BUILDING_UPGRADE_5B) {
+      this->numCreaturesInDwelling[dwellingIdx] = 0;
```
jkoppel: So Cyclops an change into Black Dragons, but Black Dragons change into nothing?

Kert: No. Cyclops will change to Green Dragons. Black Dragons will change to Cyclops
The algorithm will never upgrade creatures, only downgrade

Kert:
> Barbarbarian with Tavern to Necromancer and back
Tavern is not built anymore
> Warlock with Black Tower to Sorceress and back.
Black Tower is not built anymore
> Sorceress with everything to Warlock.
Same logic as for the others above

Kert:
> (I see the code is very explicit about this, but I still want to see it working.)

What do you want to see?

src/cpp/shared/town/town.cpp
Kert: Done

jkoppel:
Nice work!

PR #181, opened 2019-10-02
Fixed incorrect check for existence of moat

Kert:


jkoppel:
This is just a refactoring thing, right? No actual bug, right?

PR #182, opened 2019-10-07
Embed campaign metadata in savegames. New save file formats

Kert:
- New save file formats .GIC and .GCC - ironfist campaign and custom campaign
- All the info from .cmp file will be also saved to a save file
- Fixed scenarioDifficulty loading incorrectly
- Refactoring of xml related code to avoid duplication of functions
- Removed obsolete map_xml files
- Removed hover item from campaign xml


src/cpp/shared/campaign/campaign.cpp
```diff
@@ -161,12 +161,12 @@ std::map<int, std::map<int, int>> xCampaignDifficulties = {
   {3, {{0, 0}, {1, 1}, {2, 2}, {3, 2}}}
 };
 
-std::map<int, std::map<int, std::vector<int>>> mapsToComplete;
+std::map<int, std::map<int, std::set<int>>> mapsToComplete;
```
jkoppel: @Kert  Can you explain the problem this commit fixes?

jkoppel:
I want the stuff for saving campaign metadata moved alongside the stuff for loading it. (And thence separated out from the stuff for saving individual maps.)

Kert:
> I want the stuff for saving campaign metadata moved alongside the stuff for loading it. (And thence separated out from the stuff for saving individual maps.)

Done. It increased the length of many functions though.

src/cpp/shared/campaign/campaign.cpp
Kert: You can't have same values in set, so identical values won't pile up as they would do in vector. There are no clear functions used for any of those containers in our code.

src/cpp/shared/xml/utils_xml.h
```diff
@@ -1,16 +1,92 @@
...
-
+  template<typename T>
+  void PushBack(tinyxml2::XMLDocument *doc, tinyxml2::XMLNode *dest, const char* name, const T &val) {
+    tinyxml2::XMLElement *elem = doc->NewElement(name);
+    elem->SetText(val);
+    dest->InsertEndChild(elem);
+  }
+  template <typename T>
+  void WriteArray(tinyxml2::XMLDocument *doc, tinyxml2::XMLNode *dest, const char* name, const T &src) {
+    for(int i = 0; i < ELEMENTS_IN(src); i++) {
```
jkoppel: Are you sure that ELEMENTS_IN will work here?

jkoppel: It requires that the array's size is statically known.

src/cpp/shared/xml/utils_xml.h
```diff
@@ -1,16 +1,92 @@
...
+      for(auto val : i.second) {
+        tinyxml2::XMLElement *elem = doc->NewElement(name);
+        elem->SetAttribute("index", i.first);
+        elem->SetAttribute("value", val);
+        dest->InsertEndChild(elem);
+      }
+    }
+  }
+  template<typename T>
+  void WriteCampaignDDArray(tinyxml2::XMLDocument *doc, tinyxml2::XMLNode *dest, const char* name, const T &src) {
```
jkoppel: This function is specifically tied to representation choices of the campaign, and should be grouped with that code.

jkoppel:
About ready to merge. I left a couple small comments.

src/cpp/shared/campaign/campaign.cpp
jkoppel: Why could it have identical values before?

jkoppel: Oh, I missed your last sentence.

src/cpp/shared/xml/utils_xml.h
Kert: Compiler makes functions from this template for each array type like char[8], short[11] e.t.c.

src/cpp/shared/xml/utils_xml.h
Kert: The point is - it's not related to .cmp file in any way
On the other side all of the code in campaign_xml.cpp is related to .cmp file

src/cpp/shared/xml/utils_xml.h
jkoppel: template<typename T, size_t N>  void foo(const T (&t)[N]) { ... }

Declare it like this. Then you can't pass in something that's not an array here.

src/cpp/shared/xml/utils_xml.h
jkoppel: I see your point.

What I'm saying is that all the other functions in this file are general-purpose, whereas this is special purpose, so this is not a good place for it.

Kert: Moved it back to where it was

src/cpp/shared/xml/utils_xml.h
Kert: Fixed here and also in WriteCampaignDDArray

PR #183, opened 2019-10-07
Linear path for cyborg campaign scenarios

Kert:
New layout and updated graphics
xtrack


jkoppel:
Though Map 5 (CAMP5_06.HXC)  is not in a playable state, please add it to the campaign and update this metadata + installer so that we can have fully a functioning breadcrumbs to send to art.

Kert:
> Though Map 5 (CAMP5_06.HXC) is not in a playable state, please add it to the campaign and update this metadata + installer so that we can have fully a functioning breadcrumbs to send to art.

I will add it to map prototype branch

jkoppel:
Woah, nice! Do you have sources?

If so, please add them to Dropbox and update the file list in https://docs.google.com/spreadsheets/d/1y-gvrztkAMGj11rl_c7oRfU-BhxDh67baKPgyqE7o2Y/edit#gid=0 . If not, then don't worry.

Once you have done this, you may merge into the parent branch (but not master). If that's not relevant, then you may merge into the parent branch immediately.

PR #185, opened 2019-10-21
Read editor overlays from XML

Kert:


src/cpp/shared/adventure/map.h
```diff
@@ -325,6 +325,7 @@ enum ADVENTURE_MAP_LOCATION {
   LOCATION_RANDOM_MONSTER_VERY_STRONG = 54,
   LOCATION_RANDOM_HERO = 55,
   LOCATION_NOTHING_SPECIAL = 56,
+  LOCATION_WHAT_IS_THIS = 57,
```
jkoppel: Is this used?

jkoppel:
Please confirm that the overlays array obtained from doing this is binary-identical to the original one.

Once you have done so, you may merge.

src/cpp/shared/adventure/map.h
Kert: It is used in overlays array

jkoppel:
What overlay uses LOCATION_WHAT_IS_THIS?

Kert:
lol. There is a literally a screenshot of objects using this type above in this PR

jkoppel:
Oh oh oh oh oh oh.

jkoppel:
How about LOCATION_HOLE_LIKE ?

Kert:
> Please confirm that the overlays array obtained from doing this is binary-identical to the original one.
> 
> Once you have done so, you may merge.

Checked. Original array and the one read from XML are binary-identical

PR #190, opened 2019-12-18
Map 4 encounters

Kert:
Desert tents as excavation sites with battles
Messages on specific dates
Random pick of excavation site
Map win logic


jkoppel:
The game actually froze my first time trying to end the turn (in debug mode). I completed the test without ending the turn, but this is still concerning. Reason to up-prioritize investigating this.

This looks excellent. Works as advertised.

Have not tested that save/reload works correctly. I'll take your word for it if it does.

Is it intended that the message only appears once total? You said "on specific dates," but the code gives a single specific date.

Artifact 104 is the Iron Fist, rigiht? Is there no ARTIFACT_IRON_FIST constant?  Why not?

Please add a magic eye overlooking the southwest excavation site.

Kert:
eye

Kert:
@jkoppel The messages on specific dates are made through the editor.

jkoppel:
Excellent work.

PR #193, opened 2020-01-22
Plasma Cone spell

Kert:
"Creates a cone of hot plasma in the direction of the target"
This branch also:
- Changes behaviour of hex dimming when hovering during spell cast - now you see which creatures will be affected and you will see the area affected when doing fireballs/blasts/plasma cone!
- Fixed support for colored versions of current new and future spell cursors


src/cpp/shared/game/game.cpp
```diff
@@ -2220,4 +2220,120 @@ void game::ShowMoraleInfo(hero *hro, int dialogType) {
     strcat(gText, "\nnone");
 
   return NormalDialog(gText, dialogType, -1, -1, -1, 0, -1, 0, -1, 0);
+}
+
+void mouseManager::SetPointer(int spriteIdxArg) {
```
jkoppel: What changes did you make to this from the original decompiled?

Can you make it more clear?

src/cpp/shared/combat/combat_spells.cpp
```diff
@@ -632,4 +665,459 @@ void combatManager::CastSpell(int proto_spell, int hexIdx, int isCreatureAbility
...
+  
+  offsetY = screenY - 63 - 42 * (hex / 13) - 26;
+
+  int clockHour = 0;
+  int v15 = offsetX - 22;
+  if(v15 >= 0) {
+    if(offsetY >= 0)
+      clockHour = 6;
+  } else if(offsetY >= 0) {
+    clockHour = 12;
```
jkoppel: This is totally the wrong variable name, right?

src/cpp/shared/combat/combat_spells.cpp
```diff
@@ -632,4 +665,459 @@ void combatManager::CastSpell(int proto_spell, int hexIdx, int isCreatureAbility
   }
   WaitEndSample(res, res.sample);
   this->CheckChangeSelector();
+}
+
+CURSOR_DIRECTION combatManager::GetCursorDirection(int screenX, int screenY, int hex) {
```
jkoppel: Was this decompiled code?

If so, what changes did you make.

If can you do some things to make this more clear?  E.g.: Use functions to make lines 674-678 more clear.

Kert: Added a general comment on this func. I grabbed a chunk of CheckSetMouseDirection and added a switch in the end so
If you insist I'll go through the function to see how it actually manages to calculate these so-called "clock hours" and refactor it

src/cpp/shared/combat/combat_spells.cpp
Kert: which? clockHour is correct

src/cpp/shared/game/game.cpp
Kert: Changed a comment inside it. This is the only thing that changed from original decompiled afair

src/cpp/shared/combat/combat_spells.cpp
jkoppel: Oh. I see. Maybe rename it to "directionInClockHours". "clockHours" sounded to me like  a unit of time, which is  weird to have here.

jkoppel:
Change the variable name, then merge this.

src/cpp/shared/combat/combat_spells.cpp
Kert: Somehow I thought it's hours in clock (wtf clock with 24 hours???), in reality it's just 24 parts of hex
Not sure why it's divided in this many in the algo, instead of just 6
Anyway fixed

PR #196, opened 2020-02-18
Fixed showing -2 damage when hydras attack shadow assasins and they e…

Kert:
…vade the attack

### DIFF
--- a/src/cpp/shared/combat/army.cpp
+++ b/src/cpp/shared/combat/army.cpp
@@ -136,7 +136,15 @@ void DoAttackBattleMessage(army *attacker, army *target, int creaturesKilled, in
     attackingCreature = GetCreatureName(attacker->creatureIdx);
   else
     attackingCreature = GetCreaturePluralName(attacker->creatureIdx);
-  if (creaturesKilled <= 1)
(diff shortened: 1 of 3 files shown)


src/cpp/shared/combat/army.cpp
```diff
@@ -589,6 +597,69 @@ void army::DoAttack(int isRetaliation) {
 
 }
 
+void army::DoHydraAttack(int isRetaliation) {
```
jkoppel: Can you add a comment about what you changed from the original code?

Also, please explain the bug.

Kert: added a comment
It was just a problem with displaying combat message

jkoppel:
How can a hydra do < 0 damage?

Kert:
> How can a hydra do < 0 damage?

Attack a mirror image and destroy it (-1 damage) or attack assasin and he evades (-2 damage)

jkoppel:
Why are those negative damage? Where is this set?

Kert:
> Why are those negative damage? Where is this set?

army::DamageEnemy

PR #197, opened 2020-02-26
Fire Bomb spell

Kert:
"Set an area on fire for 2 turns. Creatures will continue burning for 2 turns after leaving the area"
Deals 20+-5 damage


Kert:
There's a minor bug when player hero blinks for 1 frame if you hover a cursor over something on the horizontal line where the hero is, creature info is ON and there are fire walls on the same line, spent a couple of hours and still can't fix it.

jkoppel:
Nice work @Kert

src/cpp/shared/expansions.h
```diff
@@ -8,9 +8,20 @@ class stackExtra {
...
+        int currentFrame;
+    };
+    public:
+        std::vector<FireBombWallHex> fireBombWalls;
+};
+
 class combatExtra {
     public:
         stackExtra stack;
+        spellExtra spell;
```
jkoppel: :)

PR #199, opened 2020-03-03
Fixed black'n'white cursor not being applied

Kert:


jkoppel:
It's a marvel of bad API design that this (a) compiled and (b) didn't yield a runtime error.

PR #201, opened 2020-03-11
Shipyard Object

Kert:
Currently uses dummy Standing Stones graphics
Boat cost is the same as if you buy it in a town


src/cpp/shared/game/game.cpp
```diff
@@ -1043,6 +1031,13 @@ int philAI::EvaluateHeroEvent(int a1, int a2, int a3, int a4, int *a5) {
 	return EvaluateHeroEvent_orig(a1, a2, a3, a4, a5);
 }
 
+int philAI::ValueOfEventAtPosition(int x, int y, int a2, int *a3) {
+  if(gpAdvManager->GetCell(x, y)->getLocationType() == LOCATION_SHIPYARD)
+    return 0; // ignore shipyard objects
```
jkoppel: I'm curious: Did you get the infamous "Tell Phil" popup before you fixed this?

src/cpp/shared/adventure/adv.cpp
```diff
@@ -326,6 +326,49 @@ void advManager::DoEvent(class mapCell *cell, int locX, int locY) {
...
+        if(!giGroundToTerrain[cell->groundIndex] && !cell->objType) {
+          boatPossible = true;
+          break;
+        }
+      }
+
+      if(gpGame->GetBoatsBuilt() >= MAX_BOATS || !boatPossible) {
+        NormalDialog("Cannot build another boat.", 1, 208, 40, -1, 0, -1, 0, -1, 0);
+      } else {        
+        heroWindow* wind = new heroWindow(177, 20, "shipwind.bin");
```
jkoppel: Reusing the GUI from the in-town dialog? Nice!

src/cpp/shared/game/game.cpp
jkoppel: Oh hey, I see you already posted that to the Trello card. Called it!

I got this one way back in 2010 or 2011, back when the project was just getting started.

jkoppel:
1) Can you select an appropriate sound and play it upon construction of the boat?
2) Is there an easy way to make it fizzle in?

jkoppel:
ATM, cannot enter the stones from the right. When we get the real graphics, this will be changed.

Kert:
Entering only one tile is intended (I made it 1 instead of 2). I think it was the same in homm3 - visually shipyard is big but only one tile is interactable

Kert:
> 1. Can you select an appropriate sound and play it upon construction of the boat?
> 2. Is there an easy way to make it fizzle in?

1. I can't find a fitting one. All sound way too out of place
2. Yes

PR #202, opened 2020-03-12
Pixel-perfect resolution options + disabled resizing using mouse

Kert:


jkoppel:
Wow. This wasn't so bad.

PR #206, opened 2020-05-12
Campaign Map 1

Kert:


jkoppel:
@Kert What was the status of this?

Kert:
> @Kert What was the status of this?

Not sure. Trello card should be the source of the truth I guess

jkoppel:
I'm merging this because there's no reason not to. Will make it easier to test the map with the latest changes.

PR #207, opened 2020-06-17
Inclined to join flag fix

Kert:
Inlcined to join flag will always work except when there's no place in hero's army

### DIFF
--- a/src/cpp/shared/adventure/adv.cpp
+++ b/src/cpp/shared/adventure/adv.cpp
@@ -696,7 +696,7 @@ void advManager::PlayerMonsterInteract(mapCell *monsterCell, mapCell *anotherCel
 		ScriptCallback("OnMonsterInteract", x, y);
 	}
 
-  bool joinCellFlag = monsterCell->extraInfo & (1 << 12);
+  bool joinCellFlag = monsterCell->extraInfo & MAP_CELL_EXTRA_MONSTER_AMOUNT;
   int creatureIdx = monsterCell->objectIndex;
   if(!joinCellFlag || !hro->army.CanJoin(creatureIdx) || gbInCampaign || (xIsPlayingExpansionCampaign && xCampaign.campaignID <= 3)) {
     this->PlayerMonsterInteract_orig(monsterCell, anotherCell, hro, window, x1, y1, unused, x2, y2);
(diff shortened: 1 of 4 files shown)


src/cpp/shared/adventure/adv.cpp
```diff
@@ -683,17 +685,52 @@ void advManager::QuickInfo(int x, int y) {
...
 	}
-	if (GetMapCellXY(cell, &x, &y)) {
+
+  int x, y;
+	if (GetMapCellXY(monsterCell, &x, &y)) {
 		ScriptCallback("OnMonsterInteract", x, y);
 	}
-	this->PlayerMonsterInteract_orig(cell, other, player, window, a1, a2, a3, a4, a5);
+
+  bool joinCellFlag = monsterCell->extraInfo & (1 << 12);
```
jkoppel, line 699: That 12 is a magic number and needs to be declared somewhere.

Where's the code for setting  this flag? It should already exist?

src/cpp/shared/adventure/adv.cpp
```diff
@@ -683,17 +685,52 @@ void advManager::QuickInfo(int x, int y) {
...
+  int creatureIdx = monsterCell->objectIndex;
+  if(!joinCellFlag || !hro->army.CanJoin(creatureIdx) || gbInCampaign || (xIsPlayingExpansionCampaign && xCampaign.campaignID <= 3)) {
+    this->PlayerMonsterInteract_orig(monsterCell, anotherCell, hro, window, x1, y1, unused, x2, y2);
+    return;
+  }
+
+  gpMouseManager->ShowColorPointer();
+  int quantity = (unsigned char)monsterCell->extraInfo;
+  sprintf(gText, gEventText[66], GetCreaturePluralName(creatureIdx));
```
jkoppel, line 708: More magic numbers.

jkoppel:
Where is the previous code for the inclinedToJoin flag? Did it just not exist?

Kert:
> Where is the previous code for the inclinedToJoin flag? Did it just not exist?

It's all in src/cpp/shared/scripting/funcs.cpp that you can observe in Files changed. I just removed comments because they are obsolete now

src/cpp/shared/adventure/adv.cpp
Kert: Would you like me to try and make an enum for all event texts?

Kert:
Issues addressed anyway

jkoppel:
Oh right! I remember now. We believed that setting this bit would make creatures inclined to join. However, the actual event logic was more complicated, and so it wouldn't always work.

jkoppel:
Nice work!

PR #214, opened 2020-06-30
Fixed Right-clicking on water objects/grass object category in editor

Kert:


jkoppel:
So.....the editor was just always wrong? Any idea how this bug happened?

PR #218, opened 2020-09-16
Disallow building support

Kert:
Added engine support for disallowing certain buildings in towns to be built
Mark disallowed buildings with different color on Build screen
DisallowBuilding(townIdx, building) scripting func
Scripting documentation updated


src/cpp/shared/town/town.cpp
```diff
@@ -1113,10 +1121,22 @@ void townManager::SetupExtraStuff() {
 }
 
 void townManager::SetupCastle(heroWindow *window, int a3) {
-  
+  enum BUILDING_STATUS_IMAGES {
```
jkoppel, line 1124: I like that you did this.

jkoppel:
Nice work. All that's left is to change the message upon viewing a disabled building.

PR #220, opened 2021-01-27
Fixed black&white cursors not updating correctly during combat

Kert:


src/cpp/shared/game/game.cpp
```diff
@@ -2296,6 +2293,8 @@ void mouseManager::SetPointer(int spriteIdxArg) {
       int fileID = gpResourceManager->MakeId(fileNameChar, 1);
       gpResourceManager->PointToFile(fileID);
 
+      BaseFree(cColorBits[this->cursorIdx], __FILE__, __LINE__);
+      BaseFree(cAndBits[this->cursorIdx], __FILE__, __LINE__);
```
jkoppel, line 2297: Is it guaranteed there's something allocated here?

Kert: The only thing that was done to it is
memset(cColorBits, 0, sizeof(cColorBits));
at mouseManager creation

jkoppel: Then....shouldn't BaseFree crash?

Kert: Not sure, but it didn't crash in debug

jkoppel: Yeah....this isn't confidence-inspiring enough.

Basically, my issue:

(1) I currently believe BaseFree(a, __FILE__, __LINE__) has a precondition that  a is a pointer to allocated memory. BaseFree may act benignly if a is not allocated, or it may go haywire.
(2) BaseFree is being called without something making it clear that this should be satisfied

Are you familiar with the phrase **programming by coincidence?** If this works but we don't understand why, then this is an example of that.

Basically, we need to have code that works for reasons we understand.

So, can you either correct my statements above or change this to something that works for reasons we understand?

(And yes, I do realize that this is far from an ideal codebase, but I'm still choosing to avoid making it worse in this instance in this fashion.)

Kert: Fully agree. Just don't know the way to avoid that. And since it worked I left it as it is
Btw wouldn't be checking if(!cColorBits[this->cursorIdx]) enough?

jkoppel: > Btw wouldn't be checking if(!cColorBits[this->cursorIdx]) enough?

It makes me cringe, but yes, that would do it; if you implement this, you can merge this code yourself.

PR #226, opened 2023-05-11
Improved reverse engineering of certain class fields

Kert:
Updated playerData, hero, searchArray, PathfindingInfo, boat class fields


src/cpp/shared/adventure/adv.h
```diff
@@ -66,34 +67,30 @@ class armyGroup {
...
-  char field_6;
-  char field_7;
-  char field_8;
-  char field_9;
-  char name[13];
+  short aiLastHeroInteractionTurn;
+  char aiLastHeroInteractionIdx;
+  short aiLastTownInteractionTurn;
+  char aiLastTownInteractionIdx;
+  char name[MAX_FILENAME_LENGTH];
```
jkoppel, line 77: Should use a different constant. Although the max hero name is the same as the max filename length, hero names are not filenames, and they could be changed to differ.

We actually teach why some things should be duplicated in my software design business. See https://www.pathsensitive.com/2018/01/the-design-of-software-is-thing-apart.html

We actually collect examples of cases where duplication is good. I'm adding this there.

https://docs.google.com/document/d/1DvYcmeuadYEHY9uw8ZWpdXmem_pykgJUDDaCkd_VGJs/edit#

Kert: I agree. Accidentally missed this

src/cpp/shared/adventure/adv.h
```diff
@@ -66,34 +67,30 @@ class armyGroup {
...
-  __int16 field_23;
-  __int16 field_25;
-  __int16 field_27;
-  __int8 relatedToX;
-  __int8 relatedToY;
-  __int8 relatedToFactionID;
-  __int8 directionFacing;
-  __int16 occupiedObjType;
-  __int16 occupiedObjVal;
+  int aiTargetX;
```
jkoppel, line 82: These are 32-bit ints? For an x-coordinate? Any idea why? Usually map coordinates are smaller.

jkoppel:
Nice work! Have you tested that saving, loading, then saving produces identical files?

Kert:
Yes. They are identical

